PR #613, opened 2013-09-06
Added OsmAnd (Maps), Fixed some missing translations

caffeinewriter:
http://osmand.net


RiQuY:
Can't find OsmAnd in F-Droid is it possible to run it without Google services like Google Play?

Hillside502:
It's on Amazon
https://www.amazon.com/gp/product/B00D0SEGMC/ref=mas_pm_OsmAnd-Maps-Navigation

but I don't know whether it needs gapps.

yegortimoshenko:
https://f-droid.org/en/packages/net.osmand.plus/

Hillside502:
The direct link is good, but F-Droid's search box gives 0 result.

RiQuY:
Do you have the repository to add to F-Droid? Thank you in advance.

Hillside502:
The app is in the main F-Droid repo; it's just the website search that fails.

RiQuY:
Should I download the app from F-Droid even if it is not the official upload?

Zegnat:
> Should I download the app from F-Droid even if it is not the official upload?

Do you trust F-Droid to compile apps for you? If you are already using F-Droid, the answer to that seems to be yes. In that case, why would you not trust them with OsmAnd?

The F-Droid [repository for OsmAnd](https://gitlab.com/f-droid-mirrors/OsmAnd-submodules) is actually just links to separate official GitHub repositories from OsmAnd. So there should be no source differences between an “official upload” and the F-Droid one.

Hillside502:
@RiQuY 
You're confusing matters by editing previous comments, instead of commenting afresh.

You slotted in https://github.com/osmandapp/Osmand/pull/4348 retrospectively.

PR #905, opened 2014-03-16
Added Mailpile to Email Clients in en-projects.json and logos

bnvk:
Passed build, previewed in browser, all seems to be good to go :)


pravi:
mailpile is in beta now and it works pretty well. I think it may be a good time to suggest it.

alerque:
I agree it's probably time to include Mailphile. Anybody care to work on an actual PR to submit?

PR #979, opened 2014-05-11
Add OpenConnect VPN clients

cernekee:


strugee:
Thanks so much for your contribution to PRISM Break!

I'm going to go ahead and close this because I was just going through old PRs and looked into OpenConnect, and I don't really see any compelling advantage over any existing VPN solution we recommend. In addition the protocol that OpenConnect implements seems to be Cisco proprietary, which is kinda yucky.

If I missed anything, please feel free to bring it up and we can revisit this.

Issue #998, opened 2014-05-29
AxCrypt


alex-fischer-70:
There is an interesting encryption program for the Windows platform. Although Windows can't be trusted, in some times we don't have an opprtunity to change the os. So, the tool I'm talking about is AxCrypt from Axantum.
It provides AES128 (AES256 is planned) encryption for single files with password and/or keyfile. I think this is an interesting alternative to container encryption programs. Of course it is open source software ;).
http://www.axantum.com/AxCrypt/Default.html
So, what's about AxCrypt.
Can you add this program to the list? I'm using it since 2 years and it does a goot job.

Alex

pstashp:
Symantec Anti-Virus has this feature called "Tamper Protection", which will log processes that try to mess around with Symantec processes.  AxCrypt is the only application that I see being logged as trying to temper with anti-virus.  This is really suspicious.

Looking at the instructions on how to build AxCrypt from source, it seems to use one 3rd party component and the instruction just says "just follow the instructions".  Another suspicious item.

I would love to audit the code, but I'm no cryptographer so my efforts will be futile.

I think the least I could do next is to see whether Axcrypt is phoning phone by logging any traffic from AxCrypt.

yegortimoshenko:
Subscription-based: https://www.axcrypt.net/pricing/

Some functionality (incl. 256-bit AES) is implemented on the server for the sole purpose of limiting access to users that have downloaded free version:

https://www.axcrypt.net/support/faq/

>  Our intention is not for you to pay an annual fee for the software as such, the annual fee is primarily for the server-based functionality such as Key sharing, Password manager, Secured Folders, 256-bit encryption etc. Also, we’re removing all annoying advertising and bundle offers during the installation for 2.x. In addition you’re getting something that was heavily requested in 1.x, for free, which is a true standalone portable version of AxCrypt that does not require any installation.

It's not OK to suggest encryption software that requires remote server to run.

alerque:
No objections for closing this here. I don't think we can send people this way in good conscience.

yegortimoshenko:
Good, then closing based on consensus.

PR #1000, opened 2014-05-30
Added AxCrypt for Windows platform

alex-fischer-70:


yegortimoshenko:
See https://github.com/nylira/prism-break/issues/998#issuecomment-371980937.

PR #1014, opened 2014-06-09
Add Cryptonote

halfredtook:
Cryptonote is a opensource webbased note application with client-sided encryption and synchronization with optional servers. There are currently Android, iOS, and Web clients available. So you can access your notes from any device / client supporting html and javascript.

http://www.mindbin.io


yegortimoshenko:
Site is defunct, [source code](https://github.com/Mindbin/Cryptonote) has last been updated four years ago. Safe to close.

alerque:
Agreed, this PR was premature from day one anyway.

PR #1034, opened 2014-06-22
add password manager category, add Password Safe (Windows, Linux, Android)

ThomasWaldmann:
note: I tried that copystubs.js code to add the stubs for all the other languages, but it reformatted quite a lot of unrelated stuff in the files (which is wanted IMHO, but not in the scope of this pull request)
thus, this is only for english.


TheMSB:
> Were you planning to add back 1Password in proprietary?

I am having trouble finding any discussion on this topic, if there is please direct me to it.

I was planning on writing up a PR for 1Password but from what I get from this statement is that it was removed before? Why? As a proprietary service it seems much more recommendable over LastPass since it stores the password database locally.

Zegnat:
1Password is proprietary software and will not be recommended by PRISM Break. It hasn’t been added to the proprietary list simply because nobody has done so yet. I for one use 1Password anyway, as their encryption/container format is published and can be audited to have no known backdoors. That however does not change the fact that it does not comply with PRISM Break’s guidelines.

PR #1123, opened 2014-08-29
Web Search: Searx

plastipeta:


GreenLunar:
I did not find ```searx```at https://prism-break.org/

Zegnat:
@GreenLunar this pull request hasn’t been merged yet as there has been only minimal discussion about Searx. E.g. should it replace Seeks? Replace Yacy? Maybe both?

I would like to see a little more discussion before I add yet another search tool to the list we already have.

GreenLunar:
> should it replace Seeks? Replace Yacy? Maybe both?

None, searx should be listed among both.

> I would like to see a little more discussion before I add yet another search tool to the list we already have.

Did you try it? It seems like a very good product which is also available in several languages.


P.S. My favorite and preferable search engine is Seeks.

GreenLunar:
> It distances seeks and yacy by miiiles when it comes to usability/performance/interface/features/installation.

@nodiscc what searx have that seeks does not? and how performance is different?

Zegnat:
> > should it replace Seeks? Replace Yacy? Maybe both?
> 
> None, searx should be listed among both.

PRISM Break added the following to the inclusion guidelines specifically because some categories were getting way too big (and some are still too big for my liking):

> **Quality over quantity.** PRISM Break strives to promote the best open source applications. Ease of use, stability, and performance matter. This is the first time many people are looking to leave their proprietary walled gardens. Let's make it a good experience for them.

Adding yet another search solution feels like a quantity over quality thing. That’s why I asked. If you are saying Searx is just as good as Seeks, why do you believe it should be on the list? Keeping in mind that PRISM Break is *not* a list of libre software, there are other projects that aspire to be that.

nodiscc:
For those curious about it, you can try a public instance of searx at https://searx.laquadrature.net/. I find the usability is better than Seeks. Definitely better than Yacy. 


As of today, the search engines category is not so large, so adding Searx may be a good step forward in the first time. Note: If you'd like to remove some search engines for the list, I suggest starting with non-free ones. They could be moved to the Notes section.

GreenLunar:
> If you are saying Searx is just as good as Seeks, why do you believe it should be on the list?

Because searx might add p2p communication with Seeks nodes. See https://github.com/asciimoo/searx/issues/393

wiewo:
I'm also an avid searx user. I was using other instances for over a year nearly exclusively as my only search engine and I finally set up my own instance yesterday. I did not start using it earlier because it was not listed on prism-break and I discovered it only accidentally by reading a blog post. Before I was mostly using MetaGer. When it comes to quality of results, speed, usability and look, searx easily surpasses MetaGer in every of these aspects. Especially the filter options are very handy.
I think the fair comparison would be to rank searx among the Metasearch engines like MetaGer, Unbubble and Ixquick. Although I cannot say anything about Ixquick and Unbubble because I did not use them for some time, I think searx should replace all of them if the list really needs to be reduced. Ixquick, Unbubble and MetaGer don't provide the source code to their services and it is not possible to set up your own instance. People use them solely based on trust and who knows if they are not forced by others to violate their privacy policy?

vyp:
@Zegnat 

> If you are saying Searx is just as good as Seeks, why do you believe it should be on the list?

Well not sure how relevant this is, but the last commit to Seeks was over a year ago, whereas Searx is still actively developed. That probably helps Searx's case. If one really needs to be remove from the list, I also agree that it should be a nonfree one, probably either unbubble or metager (or both>). As @nodiscc says they can still have a mention under the notes section.

GreenLunar:
I think we must not omit Seeks from list.  Listing Seeks might be a call for new contributors to Seeks.  I assume that even Searx developers would agree to this statement.

Searx is indeed a high quality meta search engine tool which also provides comfortable ways to monitor results with CSV, JSON and RSS.  Searx should be listed already.

alerque:
@GreenLunar I'm not sure that line of reasoning works here. This is not an OSS recruiting directory. We don't make it a point to list software that needs help. In fact we specifically list software that is in _good shape to be used by end users_, not things that are hurting for development interest. Programmers interested in the OSS scene can be expected to do their own research and make their own decisions about what to use. This directory is about helping the less savvy make informed choices. Listing software that is no longer actively developed or falling behind the curve due to lack of developers does not serve the interests of those users. By trying to add on rationals like that to serve other interests I think we would end up hurting the interests of the primary focus of the directory.

GreenLunar:
@alerque, I understand and agree with the outlook you convey.

I think search engines and social networking platforms are critical enough to make an exception for them.  I believe search engine decentralization is concerned to every one of us; for I believe that the more to use centralized search engines, the less objective and quality information be exposed to us, even to those who use decentralized search engines, not necessarily meta-search engines like Seeks and Searx.

alerque:
@GreenLunar No, we don't agree. We disagree. I'm totally fine with disagreeing and having a discussion starting from there but please don't patronize me by saying you agree when what you are proposing is the opposite of what I am proposing.

In regard to your "exception" case (which is the single main case I had in mind when writing), for the moment at least I stand by my last assessment:

> By trying to add on rationals like that to serve other interests I think we would end up hurting the interests of the primary focus of the directory.

If one OSS search engine project is inferior to another due to the state of the development process then we would do the target audience of this site a dis-service in promoting it as a software suggestion as opposed to just recommending the project that is in better shape. If it's a matter of being best in its class in spite of the development issues then it might make sense to list with a warning about the state of the project.

On the other hand if we want to promote decentralized search as important to the future of the web the place to do that would be in the notes, not by compromising the recommendations list with products that aren't useful to end users. That being said your case for the importance of such decentralized search engines seems to hinge around information bubbles, freedom of press, and that sort of thing more than they do privacy and security and surveillance drag nets. This seems like a bit of a miss-match for this site. Not that I disagree it's important, but it's a different battle. Anything we do promote in notes outside of software recommendations should be reasoned out along the lines of the primary goals of this project.

GreenLunar:
> @GreenLunar No, we don't agree.

Message is hidden with `<!-- content -->` because it is off topic.



> If one OSS search engine project is inferior to another due to the state of the development process then we would do the target audience of this site a dis-service in promoting it as a software suggestion as opposed to just recommending the project that is in better shape. If it's a matter of being best in its class in spite of the development issues then it might make sense to list with a warning about the state of the project.

Some would disagree.  Still, there are several active Seeks nodes, some are older and some were started after searx was initiated.

wiewo:
Any update on this?
Just want to add another reason to include searx. The inclusion policies make an exception only for the search engines in regards of free software:
>PRISM Break follows the GNU/FSF definition of Free Software and prefers software licensed under a compatible license but may allow other OSI reviewed licenses. The only exception is when free software offers no viable alternative to proprietary software. "Web Search" is the only category with this exception currently.

So if PRISM Break wants to get rid of this exception, viable free software alternatives like searx should be included in the search category. With Seeks and Yacy, there would be three free software search engines in the category and the exception could be removed, if the rest gets moved to the notes section.

strugee:
@wiewo just because three free software projects exist doesn't mean that three viable free software projects exist.

wiewo:
@strugee I don't understand why these three free software projects are not at least considered viable.
Searx has more sources for aggregation of search results than all the other listed meta search engines, except for DuckDuckGo, but searx includes results from DuckDuckGo. Metager has some that are not included in Searx, but they are likely only relevant for German-speaking users or if you are searching for some specific scientific topics. Ixquick does not even share information about the sources they use, but I am not aware of anything that Ixquick brings to the table that others do not. Same for Unbubble.
Startpage is only a proxy for google search results. AFAIK Seeks can provide you with that, too. And Seeks has even more features. Or you can configure Searx in a matter of seconds to only provide results from google.
So imho Seeks and Searx alone can replace all other listed services that do not provide all their source code as free software.
But I even consider Yacy as the optimal solution. Yacy does not rely on any third party service, is a secure and anonymous distributed system and even has its own crawler. It may not satisfy the average Joe with all of its search results (yet), but if you invest some time and crawl some pages where you regularly need results from, you can get better results than with any of the other available search engines.

GreenLunar:
> @wiewo just because three free software projects exist doesn't mean that three viable free software projects exist.

This statement is correct, but all of the above are useful indeed, and are much more preferable as @wiewo has wrote in his last post on this thread.

nodiscc:
@nylira @Zegnat Can this be merged?

yegortimoshenko:
Please merge this. Searx is the only usable, fully free search engine, unlike all other options listed on PRISM Break that are either proprietary or useless. It's also very widely deployed: https://github.com/asciimoo/searx/wiki/Searx-instances

Even Riseup has a Searx instance: https://searx.riseup.net

/cc @alerque @strugee

yegortimoshenko:
If this is not mergeable in its current state for whatever reason, I can rebase it on top of master.

RiQuY:
Any news on adding Searx?

yegortimoshenko:
As I've argued elsewhere, I think Searx is the only viable free/hostable search engine: https://github.com/nylira/prism-break/pull/1890#issue-172508438 Would like to go ahead and merge this, unless anyone feels otherwise.

Issue #552, opened 2013-08-22
TextSecure and RedPhone rely on proprietary services


exteroceptiveunfastening:
It's contradictory and unsafe to recommend TextSecure and RedPhone even though they are open source software because they rely on the existence of APIs provided by the Play Services Framework. These apps also have no distribution method other than the proprietary Google Play Store. 

This issue has been addressed by Moxie and he has no plans to change this as of 3 months ago. Link here:
https://github.com/WhisperSystems/RedPhone/issues/143

Dyras:
WhisperSystems has now removed Signal's dependency on Play Store. Maybe it's time to reconsider adding it?

strugee:
Yes, but I'd like to keep discussion in one place. Our discussions about Signal have gotten really scattered across different issues over the year. I'd like to see someone put together a canonical list of things that are good and bad about Signal that've been brought up in the past - I'll do this at some point when I have time but anyone's welcome to take this on in the meantime

PR #1127, opened 2014-08-29
Added SignaliOS

plastipeta:


TheMSB:
I'm just curious after seeing the discussion going on in #1239 and #552 
From what I understand of OpenWhisperSystem's software is that Signal is pretty much a merger of TextSecure and Redphone but then on iOS. The mentioned apps were both excluded from PRISM-break's list because they rely on proprietary software(play services).

Since I am not an expert in OpenWhisperSystem's software, even if Singal on iOS uses proprietary software from Apple to deliver push notifications, wouldn't the use of Signal still be better than using plains SMS and phone calls? Seeing as you already 'trust' iOS.
(just trying to get a better understanding here)

vyp:
@TheMSB I guess the potential for danger here is if the push notifications have access to the content (i.e. text message content)? I don't suppose it would because that sort of defeats the purpose of end to end encryption I think. So if none of the services provided by the OS have access to the actual content, yes I do think it should be better.

The key difference between iOS and android is that iOS is proprietary, so one point or another in the stack, you're going to need to trust iOS anyway (almost making using iOS a lost cause for privacy for me). And so the reason to reject the usage of proprietary play services in android apps is that we know we can do better, but with iOS we cannot really.

Zegnat:
> And so the reason to reject the usage of proprietary play services in android apps is that we know we can do better, but with iOS we cannot really.

Another big reason is that Android applications that make use of proprietary libraries (and/or Google services) might not work on alternative operating systems like Replicant. PRISM Break recommends people use these to avoid possible security holes already, so recommending software that doesn’t run on those OS at all would be weird.

TheMSB:
Are there any objections left to this PR or #1392 (the iOS portion at least)?

Sigmun:
What I don't understand on this discussion is that it seems to be a problem that Signal use iOS which is not trustable so it cannot be recommended. But then, why TextSecure for instance is recommended ?

strugee:
@Sigmun ChatSecure is also (was also?) available on F-Droid on Android.

Issue #633, opened 2013-09-11
Add "Virtualization" category


sochotnicky:
I believe a lot of people (even on *NIX systems) are stuck with VMWare and similar less-than ideal solutions simply because they don't know there are alternatives. 

First suggestion would be
# proprietary
- VMware
# Free/Open Source
- KVM
- XEN
- DOSBox (for old games maybe?)
- VirtualBox 

I guess some might argue that VirtualBox is open source and while core is, several important features are only available in proprietary extension. Problem might be that above are *NIX (or even Linux) only solutions with exception of VirtualBox so it is "good enough" in this sense if people are stuck on Mac OS or Windows.

Looks reasonable?

Atavic:
[GNS3](https://github.com/GNS3/gns3-gui) Graphical Network Simulator.

alerque:
GNS has seemly nothing to do with this issue @Atavic.

As for a `Virtualization` category I'd be in favor, but I think only point and click solutions that would be direct matches for things like VMware should be considered. More knowledgeable folks know when to apply KVM,  Xen, etc. The group we need to catch is the ones that need VirtualBox. I think we can probably recommend that (with a caveat about the proprietary add-ons module).

strugee:
I would question whether we need a "virtualization" category. Is there any evidence that these contribute to PRISM? In the past we have declined to list e.g. LibreOffice as an alternative to Microsoft Office or Apple iWork (IIRC) because there wasn't a ton of evidence that just running those offline was problematic.

strugee:
See https://github.com/nylira/prism-break/issues/834#issuecomment-32602419

alerque:
@strugee This is an interesting topic and one that might need revisiting. VMWare does have a deeply ingrained habit of "phoning home". It's not quite impossible to keep it running firewalled in a corner, but it's not easy. Given a link to the net it also seems to do a lot more than just check it's license status.

By the same argument it might be time to review our long standing ignoring of office suites as well. When we first started ignoring them it was generally possible to just let them hum along offline. However with recent versions increasingly sharing document and state data to their respective cloud counterparts this argument is becoming less and less valid. The amount of surveillance that could be done on a user if you could watch what their office suite is doing online is pretty significant. The lines between online suites like Google Docs and offline suites like MS Office and iWork are pretty blurry at this point.

[...]

PR #1130, opened 2014-08-29
Added GNUCash

plastipeta:


yegortimoshenko:
Not sure what the proprietary spying alternative here would be. PRISM Break is privacy-oriented, and as such only lists proprietary software that has a potential for backdoors or spying. This has been established multiple times by @Zegnat, and it makes a lot of sense, otherwise PRISM Break would become just another free software directory.

alerque:
As I recently [argued for a Virtualization category](https://github.com/nylira/prism-break/issues/633#issuecomment-396829549) and also mentioned the office suite topic, it is actually increasingly difficult to find financial management software that stays offline or responsibly handles data storage. While 5 or 10 years ago it was true there were many popular proprietary offline apps for this, the selection pool is dwindling. It might be time to start surfacing this as an issue for genres of apps that previously were mostly offline anyway but are not making none-too-safe forays into "cloud" space.

Issue #1104, opened 2014-08-16
LocalBox


helgafirer:
Two new open source alternatives to commercial and proprietary file hosting / syncing:
1. Syncthing (http://syncthing.net/)
2. LocalBox (http://www.libbit.eu/en/producten-en/localbox)

yegortimoshenko:
LocalBox has moved over to http://www.yourlocalbox.org/, source code is at https://github.com/yourlocalbox?tab=repositories. This is not a community project or effort, but driven by a company, nothing has not been updated since 1 March 2017. Syncthing is so much better that this seems pretty safe to close.

strugee:
https://prism-break.org/en/all/#file-storage-sync is also somewhat overloaded already.

PR #1132, opened 2014-08-29
Added LocalBox

plastipeta:


strugee:
Hey, so sorry this never got dealt with. I've just closed #1104, so I'm closing this too - see that issue for the reasons we're not listing this. Thanks for contributing to PRISM Break!

PR #1135, opened 2014-08-29
Added miniLock

plastipeta:


yegortimoshenko:
Chrome only: http://minilock.io/

Issue #1156, opened 2014-10-03
APG vs. OpenKeychain


dschuermann:
I read that you want to have quality over quantity, so I would propose removing APG as a suggested PGP app for Android. It has not been updated in a while. Sure I am biased because I am the maintainer of OpenKeychain, but also new versions of K-9 Mail will no longer contain the interface to APG (see https://github.com/k9mail/k-9/pull/499)

Sephula909:
I've used both, and prefer OpenKeychain, myself. No advantage, but potential disadvantages with APG. OpenKeychain is definitely better from a user perspective, too.

qw3ry:
Not sure why this is still open, as #1171 and #1581 have been merged; if this is still open for discussion, I just noticed that APG got updated again. Not sure about re-adding it as a recommendation though, as I don't use it myself.

dschuermann:
I was able to close this as I am the original reporter.

@qw3ry where do you see an update to APG?

qw3ry:
@dschuermann for me the Google play store page says 

> A fork of APG called OpenKeychain added most of the new features, which have now been merged back into APG.
APG development is active again and I'm mostly working on the implementation of PGP/MIME now.

I've not further investigated this, but I was told it would work with K9 again

dschuermann:
@qw3ry Ah okay :) The text on the Google Play store page is very old. He never worked on PGP/MIME for APG.

PR #1171, opened 2014-10-13
remove APG recomendation per #1156

alerque:
In the interest of keeping the suggestion list uncluttered in favor of
the "best in class" products, remove the suggestion for use of APG as an
OpenPGP client in favor of just OpenKeychain. Discussion can be found on
Github issue #1156, but a summary of the case for this is that:
- some interfaces needed to play along with other recommendations such as
  K-9 Mail have been deprecated and it is no longer capable of doing the job in
  conjunction with other PRISM-Break recommendations
- the interface is much more complex than the alternative
  recommendation, does not conform to the Android interface guidelines
  well and thus and poses a significant usability hurdle to new users
- AGP isn't receiving regular developer attention

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ If you want to edit or add to the project logos on this site, look here:
     ./source/assets/images/logos/original/        # for high res/svg versions
 
 ### Add the Project Data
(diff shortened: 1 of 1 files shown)


strugee:
@alerque looks like this never happened?

dschuermann:
Can be closed. This has now been done in https://github.com/nylira/prism-break/pull/1681

strugee:
Weird. https://prism-break.org/en/projects/apg/ is still on the site...

PR #1172, opened 2014-10-13
Add Miranda NG

tweimer:
Hi,
I have added Miranda NG to the Instant Messaging clients. It supports the protocols Tox, IRC and XMPP, and has support for OTR and GPG via plugins. Tell me if you want to merge it please.


tweimer:
So, is anyone still interested in accepting this pull request? If no, what do I have to change?

I just updated the links to Miranda NG

yegortimoshenko:
Pidgin, which is architecturally similar to Miranda has been recently removed from PRISM Break. Also, the only non-proprietary protocol Miranda [supports](https://github.com/miranda-ng/miranda-ng#protocol-support) is XMPP (Jabber).

yegortimoshenko:
You are right, sorry! But note on similarity to Pidgin (in supporting as many protocols as possible) still stays.

tweimer:
Well, it would be interesting to know why Pidgin has been removed. Just because it is similar to Pidgin doesn't mean that the same reasons apply to Miranda NG as well.

yegortimoshenko:
Because implementing so many protocols greatly increases attack surface and *generally* reduces quality of each particular implementation. That especially applies to chats implemented in languages that are not memory safe. This is why Pidgin was removed, and why adding Miranda might not be a good idea.

Additionally, Miranda is Windows only, which might make it harder for users that adopt Miranda to switch to Linux (something that we ultimately want them to do). This is not a fatal flaw, but certainly problematic.

PR #1178, opened 2014-10-14
Added Tinfoil Chat

maqp:
I'm not sure how github reacts if I use current (non-translated) version of fi-projects.json with TFC pull request, i.e. does it require reverting the language to original, so this pull request is based on the fully translated finnish version (and thus, the translation is offered here aswell).


maqp:
@nylira 
@alerque 
@Zegnat 
Having an automatically mergeable pull-request hang for months without discussion is somewhat strange, especially when Alerque has already read the whitepaper. Not to spam - but could I get a status update?

Zegnat:
For me personally the main reason I haven’t merged this in is that PRISM Break hasn’t reached any sort of consensus yet when it comes to hardware solutions. They have been discussed a few times, and the general feel has always been that PRISM Break is about software only. If this were to change I would definitely get TFC in.

It doesn’t help that the PRISM Break community isn’t big, as this leads to discussions dying off quickly :disappointed:

maqp:
@Zegnat 
*None* of the additional hardware is needed for the software to run. The system is quite modular and need for hardware depends on user's threat model:

User can use internal RNG of OS for all cases. Alternatively the HWRNG of processor, or prebuilt third party HWRNG can also be used. The latter tools boost the security of almost any tool that can support them.

If the threat model is not end-point exploitation, the OTP encrypted messages renders collect-it-all dragnet surveillance provably useless; OTP is still used with numbers stations etc. - when certainty is required. Another benefit of both TFC versions compared to OTR is no risk of man-in-the-middle attack.

If the end-point exploitation is part of the threat model (of which we have clear evidence), there are currently *no* alternatives available. It is not absolutely mandatory for user to build the data diode in this case either. Even just cutting the serial cable's wire - used to send data to "wrong" direction - brings users much more security than if they were using a single computer. I also intend to add support for high-speed data diode that can be built from fiber to ethernet converters, to allow users to create a high assurance system without having to even touch soldering iron.

So to sum it up, TFC is more secure than OTR, even when used only as software with ordinary devices - and it is the only tool that scales up with hardware - to stand a chance against APTs and their zero days.

yegortimoshenko:
This requires very complex setup and doesn't provide additional security for most threat models. This also is not very actively maintained (22 commits, last at 30 Aug 2017). I would like to vote against merging this.

maqp:
@yegortimoshenko You shouldn't be counting the commits. There's [two ways to build secure systems](https://www.youtube.com/watch?v=uUXV2YPLtFc#t=20m23s):

1. "Get it right the first time". That would be things like airplanes, satellites etc. where you really can't push updates to the system.

2. "Security through agility". This is what most of secure apps etc. are. Apps that are built on insecure platforms, and that have to deal with the never-ending cat-and-mouse game of vulnerabilities and patches. This is great for general purpose communication, but we've [seen](https://thehackernews.com/2016/01/linux-kernel-hacker.html) zero days patched that have lived in systems for years. Considering the track record of general purpose OSs, it's extremely unlikely any up-to-date OS is secure.

TFC doesn't have to deal with the same problems because it's designed to provide endpoint security. But no software is perfect right? If I had found a vulnerability by now, I would have [immediately notified users](https://gist.github.com/maqp/038a93cbb96ae93b508014abaa2d25fe). TFC could in theory update it's NH and RxM remotely. However, TxM is a system that can not be updated because by definition it can not receive data from network. So every time users need to update, they need to get new hardware. This is unfortunate, so for every release, I need to get everything working as well as I can. This explains why you'll only see commits when I release the next version. (I also sign every release with RxM-like HSM so I don't want to manually type signatures at the end of the day.) 

As for the commits, the private developer repository has ~1,800 of those at the moment. You can find public copy of the state of the developer version [here](https://www.cs.helsinki.fi/u/oottela/tfc/). It's unlisted, because I don't want people to download an unsigned version from master every two days, and then force every peer to update because the protocol was fine-tuned.

TFC is currently undergoing a **major** change as it's moving from Pidgin to Tor v3 Onion Service back end. Unfortunately, [Tor Onion Service descriptors' revision counter has an issue](https://trac.torproject.org/projects/tor/ticket/25124) that prevents consecutive logins. Tails 3.6.2 is still using too old version of Tor so it can not host v3 Onion Services with stem. Stem (the Tor API for Python) 1.7 is currently being developed, so we need to wait that as well. TFC is also waiting for Ubuntu 18.04 LTS release and [PyNaCL 1.3](https://github.com/pyca/pynacl) that provides the XChaCha20-Poly1305 AEAD.

> This requires very complex setup 

Yes and no. Once the next version is ready, by copy-pasting a single command to terminal, you can install a `local test version` that provides more security and anonymity than what the current version of Ricochet (considered to be even more secure than Signal) does. If you want end-point security, you need the hardware. If there was a simpler way to achieve endpoint security, I would have done so already. You can read the wiki article on [security design](https://github.com/maqp/tfc/wiki/Security-design) to understand the design rationale. Quoting Wiki's FAQ

> End-to-end encryption on a networked system might be more usable, and it does increase the security of all users in a relative sense. However, unless the increased security stops the adversary, the benefit is inadequate. Security that stops determined attackers will always involve more rules, layers, and inconvenience.
>
> TFC is a step back in usability and not a guaranteed solution, but its security architecture provides a level of security no app can ever provide. That being said, TFC is the most convenient tool for anyone who considers adversaries that hack endpoints part of their threat model.

It's also the case that [law enforcement hacking is becoming cheapest form of surveillance](https://media.ccc.de/v/33c3-8136-stopping_law_enforcement_hacking#t=1440) because zero-day exploit deployment cost is negligible compared to the sunk cost of its development. Quoting Soghoian

> If the government hacks enough people, hacking not only becomes an attractive way of surveilling, it becomes the cheapest way of spying on people. Two or three officers can conduct one of these operations and hack thousands, tens of thousands or hundreds of thousands of targets, and as long as the operation happens in a relatively short period of time, they'll be able to get that many people before the software industry finds out and rolls out a patch and deploys the patch. So my concern is that when they hack enough people, surveillance becomes so cheap -- it becomes cheaper than a single hour of law enforcement overtime -- that this will become the tool of first resort. Hacking will be the first tool in the toolkit that they reach for before they go undercover. Before they try to convince someone the old fashioned way. My concern is that hacking is making spying far too cheap.

(You should really watch the entire talk). I think the security industry has been largely avoiding the issue because there's no app to fix the problem. And because the public debate isn't yet ready for it. Infosec people are still battling to get people to use Signal, so they don't have time or interest to think about this problem. But major governments are not waiting. NSA's Tailored Access Operations has existed since 1998, FBI's Remote Operations Unit was established in 1999. The one thing computers are good for, is automating tasks. If these guys haven't figured out ways to break into generic operating systems in 20 years with the press of a button (provided they have a zero-day exploit available), they should probably be fired.

We already have Signal, Ricochet etc that provide strong security on networked endpoints. Yet Prism-Break takes time to feature "experimental" apps like Riot that do not provide end-to-end encryption by default. I mean, Facebook Messenger, Google Allo and Telegram all sit in the same boat, yet the site shuns away from those. This puzzles me: A decentralized back end that has access to plaintexts by default is OK, but something that is aimed at people who face the most dangerous threat model can't have the best security architecture currently available? Why would you decide what's too complex for them? Why would you feature dozens of apps the differences of which are more cosmetic than architectural?

maqp:
On a side note, regardless of the final decision, this PR should wait until the next release of TFC because

* The features, usability, documentation etc. are going to change and improve
* The description for the application is going to change
* Ubuntu 17.04 the currently TFC release is for, is no longer supportet. It makes no sense to add support for TFC 1.17.08 on Ubuntu 16.04 or 17.10 at this point.

yegortimoshenko:
I appreciate the detailed response! I am going to watch that talk, and read resources you've linked to, but I'll make a few quick responses first.

Pidgin was another issue I had with Tinfoil Chat, it has been recently removed from the list: https://github.com/nylira/prism-break/issues/1653

It is great that there is a less complex setup option available in the new version. 

You've changed my mind on the issue. I support merging in Tinfoil Chat once onion service backend is implemented (only because we don't recommend Adium anymore).

maqp:
@yegortimoshenko I agree with the thoughts expressed in that PR. The current Wiki recommends routing Pidgin through Onion Services, and with TFC everything is end-to-end encrypted and unlike with OTR-plugin, with TFC you can't accidentally disable the E2EE. 

But since Tails is now phasing Pidgin out I decided it was time to move on. The nice thing about next TFC is, it is always at least as secure and anonymous as Ricochet. But like you can deanonymize yourself with Tor if you're not careful, this system has ways to shoot oneself in the foot regarding end point security. So I'm working hard on the documentation etc. to make sure user knows where the pitfalls are.

yegortimoshenko:
Finally went through all the links! Closing this issue in favor of https://github.com/nylira/prism-break/pull/1952 (which is the same but rebased against recent master). @maqp, please let us know when next TFC release comes out.

maqp:
@yegortimoshenko The [new version](https://github.com/maqp/tfc/wiki/Update-Log#tfc-11901-update-log) is now released.

alerque:
@maqp Please note this issue tracker is obsolete. Please find or open a relevant issue an [the Gitlab issue tracker](https://gitlab.com/prism-break/prism-break/issues) currently being used by the project.

PR #1188, opened 2014-11-24
Add swisscows.ch to Search engines

amilopowers:
I think www.swisscows.ch is a pretty good alternative search engine which should be added to the recommendations. Their servers are located entirely in Switzerland and they don't save anything from its users. Their privacy statement: https://swisscows.ch/privacy

The search results are very good compared to Duckduckgo or startpage. Btw.: No I don't work for them.

I am sorry if I did someting wrong, I am using Github for the first time.


alerque:
Can we close this PR? The discussion on the service is inconclusive and the PR itself is horribly broken. As it's against the wrong branches it's not going to work at all. If the service is ever accepted we'll need a new PR anyway.

vyp:
Agreed.

PR #1202, opened 2014-12-16
Added Firefox Hello to Video & Voice section

aleksandar-todorovic:
Feel free to contact me if something is wrong with my pull request.


flosse:
are there any doubts to add FF-Hello?
@aleksandar-todorovic could you update the PR to avoid merge conflicts?

vyp:
Can anyone fill us in on the telefonica stuff? (i.e. why it is/isn't a problem?)

vyp:
i.e. what exactly being "powered" by telefonica means?

strugee:
It means they use Telefonica's OpenTok product. Presumably the main features Firefox Hello depends on are firewall negotiation and stuff like that, but I'm not sure.

aleksandar-todorovic:
Yes, I could update it if needed, but since it took almost 8 months for any discussion here to happen, I'm going to wait for this to be approved first.

vyp:
> It means they use Telefonica's OpenTok product.

I might be missing something but I don't see a problem with this.

strugee:
> I might be missing something but I don't see a problem with this.

I don't either, although I didn't research it enough to tell for sure. It just took me a while to get to that name, so I figured I'd write it down.

TheMSB:
@flosse If I am not mistaken Firefox Hello uses webRTC. There seem to be some concerns about how safe to protocol is in relation to privacy. See https://github.com/EFForg/privacybadgerchrome/issues/332 for example.

aleksandar-todorovic:
Okay, I'm closing this down. There's really no point in submitting something and then waiting for _8 freaking months_ before even starting a discussion. This project is as good as dead. I'm out.

real-or-random:
I think Firefox Hello should be in the list. We should continue this pull request, even though the author of the pull request is not happy with the speed of the process...

alerque:
In poking around for information on this it seems the jury is still out and a lot of doubts are raised about the privacy of the FF Hello system. Some of it is FUD of course, but I wasn't able to find satisfactory answers to some of the issues. I would suggest Prism-Break (as it generally does) take a "guilty until proved innocent" approach and not consider including this until at least a couple knowledgeable voices express some confidence in the Privacy/Security of the system (including meta data leakage) against mass surveillance.

real-or-random:
This is a nice overview:
http://security.stackexchange.com/questions/94284/how-secure-is-firefox-hello

It also refers to a nice discussion of the privacy issue: http://mozillamediagoddess.org/2015/09/10/webrtc-privacy/

I don't see a privacy issue with Firefox Hello directly.

Firefox leaks your public IP address via STUN to make WebRTC work. I agree that this is not good but this is an issue of Firefox (and not just of Firefox Hello) because any website can exploit it: https://diafygi.github.io/webrtc-ips/ So to "solve" this privacy problem, prism-break would have to stop recommending as a _browser_. (I think this is not a good idea.)  

What does this mean for Firefox Hello? Firefox Hello uses WebRTC and thus STUN. However, this is common among the included "video & voice" tools. Linphone, CSipSimple, and Jitsi all use STUN. (As far as I remember, it's even enabled by default in Jitsi).

In my opinion, the main drawback is not the privacy but the trusted signalling server. This is indeed a difference to the mentioned tools that implement ZRTP for some man-in-the-middle protection (or to Tox, which can be used to verify fingerprints as far as I remember). 

I think we have enough information to reach _some_ informed decision.

Zegnat:
Those are some really interesting articles. If an informed decision can be reached by people actually interested in mobile (i.e. not me) I have no problem merging Firefox Hello into the project if that is what gets decided!

I’ll reopen this to make those discussions more public. Thanks for the links, @real-or-random!

aleksandar-todorovic:
It's been a year since I have submitted this as a pull request. Open up a new pull request and leave this one closed. I no longer have interest in this project what so ever.

Zegnat:
@aleksandar-todorovic I would urge you to press “unsubscribe” on the right and you will not have to have anything to do with this discussion. I see no merit in splitting up the discussion when we have a perfectly fine thread here. In fact, this is _the_ discussion thread for Hello.

Trying to stick to one thread is exactly what the PRISM Break guidelines tell people to do:

> If the issue hasn't been addressed, add a comment as to why it deserves inclusion.

mimi89999:
Wasn't Hello removed from Firefox? I guess that this PR can be closed now.

Hillside502:
History of Firefox - Wikipedia, the free encyclopedia
https://en.wikipedia.org/wiki/History_of_Firefox#Version_49

PR #1227, opened 2015-01-28
Removed MediaCrush as a Web Service

jdiez17:
MediaCrush as a service shuts down today.


Zegnat:
Thank you for the heads up, merged!

PR #1232, opened 2015-02-10
MyKolab is now Kolab Now

grote:
see https://kolabnow.com/news/mykolab-receives-largest-update-ever-rebranded-kolab-now


Zegnat:
Seems fine, when I get home I’ll merge this in and probably update the other language files also.

grote:
Thanks, yes updating the other language files would be great!

Zegnat:
All should be set down, sorry for taking an extra day. I like Kolab Now as a project so I decided to deviate from my usual task and even provide new translations for both Dutch and Swedish.

Let me know if I missed anything.

PR #1257, opened 2015-03-21
Add gogs for "File Storage & Sync"

hasufell:
homepage: http://gogs.io/
source code: https://github.com/gogits/gogs

What it is:
A painless self-hosted Git service written in Go.

Reasons:
* simpler than gitlab, less feature creep, but still powerful (ldap, login via github acc, ...), see https://github.com/gogits/gogs#features
* ~10 times easier to deploy than gitlab which really isn't for the inexperienced admin (I run both)
* a lot faster than gitlab, since Gogs is written in Go... that means you can reasonably run it on a cheap vserver which is problematic with gitlab

Remarks:
* pull requests feature is WIP and will probably be finished in the near future, see https://github.com/gogits/gogs/issues/5

license is MIT


vyp:
Honestly, I still feel it is too 'raw' at the moment. Maybe wait a couple more months?

hasufell:
> Honestly, I still feel it is too 'raw' at the moment. Maybe wait a couple more months?

That's too broad for an argument. It is not in experimental stage. If you can name a crucial security-relevant feature that is missing or broken, then that would be a different case.

vyp:
I might be wrong on some or all of these, but last I checked:

- No search feature for issue tracker. Essential when number of issues becomes high so that reporters/contributors can easily find duplicates.
    - No search for repositories either (like a simple one with github for example).

- Syntax highlighting is still 'off' in any places.
    - By off, I mean that the syntax highlighting is very simple (does not take kind to regex/edge cases), also trys to highlight plain text files.
    - Also outright lacks highlighting for some filetypes (php, haskell, cabal, sql).
    - Not critical issue, but just adds to the feel of being 'raw'.

- Irregularities between the layouts between different types of pages. I think this was because they were changing the frontend layout 'engine', but only changed it for the issue tracking pages, and not the repository viewing pages. Not critical, but blah blah..

- Cannot edit comments on issue tracker.

- User's public activity is sort of weird, buggy:
    - Can show private repository activity.
    - Sometimes shows other users' activity too.

- Submodules are not shown correctly.

I think there may be a few more *somewhat* 'major' ones in the issue list. I just took this off the top of my head.

Don't get me wrong, I'm loving gogs, and I'm certainly not opposed to adding it, but I'm hesitant to list it on prism break at this state. And yes, I have seen the massive (2200+) number of commits to it.

But of course this is just my opinion.

hasufell:
prism-break is not really about detailed feature-review of software, unless those features are security-relevant

gogs is stable and usable for serious purposes... but it is not a drop-in replacement for gitlab

given the maintenance cost of gitlab, this is currently the best alternative, especially for people who just want something to easily fire up and work with... as in: perfect for the prism-break audience

vyp:
Yes I think you're right about the security aspect, which I initially missed. See my edit in previous comment.

vyp:
Also, imo pull request feature is not essential or critical or anything, so I don't think lack of that should prevent it being listed at the moment either, especially for the use case/audience of people who just want to fire it up and use it easily.

hasufell:
> imo pull request feature is not essential or critical or anything

It's pretty much the last thing missing before I remove gitlab from my server. The rest (like missing search etc) are rather small things that'll get fixed sometime anyway.

For prism-break itself (as in: self-host prism-break) it's something that's probably not mature enough in terms of features, but should be watched.

vyp:
Well I mean one can always just open a new issue, put `[PR]` in the title, and give the repository url and branch to pull from.

hasufell:
> Well I mean one can always just open a new issue, put [PR] in the title, and give the repository url and branch to pull from.

yeah sure... or just use git send-email :)

vyp:
Oh sorry, well I just saw your comment on gogs: https://github.com/gogits/gogs/issues/5#issuecomment-75776560. Guess a pull doesn't solve that.

Usually what I would do is create a new branch to pull into, and then diff and test in there before merging into a main branch (or just delete the branch otherwise). But I'm not sure if that's really the way to do it because I'm a git newbie. And also it requires that the machine has git (meaning no review on a machine without git, i.e. phones).

hasufell:
gogs now has pull-request support

PR #1258, opened 2015-03-23
Fixed couple broken links

wifiextender:
Hello again,

I have fixed couple broken links.

I was split between the iceweasel links [1](https://packages.debian.org/experimental/iceweasel#pdownload) and [2](https://wiki.debian.org/Iceweasel), because the first link could be used by other distributions too.

At the end decided that the second link will be easier to follow, as most people are casual computer users, and the first link might put them off.

Cheers


nylira:
Yeah, the second link is more user friendly. Thanks.

PR #1259, opened 2015-03-23
French mistake and more translations.

smonff:
I believe prism-break.org is very useful and a high quality software index. As I use it regularly  and found some mistakes in french translations, I tried to contribute by fixing some stuff! Hope it helps.

Also adds some new translations. 
If this work is ok, I could commit more things later.


nylira:
This looks good, thanks.

PR #1263, opened 2015-03-28
Added FileCast

onyxbits:


mastercoms:
This project seems like it is in its early stages, it was started 3 days ago and has 6 commits.

onyxbits:
It is fully functional. It has only 6 commits because I created the repo late.

mastercoms:
Ah, sorry about the confusion and presumptions, then.

Zegnat:
This looks like a really interesting and cool tool. But I am not sure what sort of proprietary software this is supposed to be replacing. Instead, it just seems to be a cool little open-source tool. @onyxbits, why do you believe this should be featured on PRISM Break? Maybe I’m missing something.

I am also a bit wary of featuring what is basically a web server when it doesn’t have HTTPS support.

onyxbits:
FileCast is meant as a replacement for services like dropbox or imgur. The whole idea is to keep files you want to transfer in your local network. A typical usecase would be a friend coming over and wanting to grab some of your files, but neither of you is particularly keen on setting up any kind of cloud syncing.

I would argue that your WLAN is probably not infiltrated, so HTTP is as safe as any network share you might have setup. Nevertheless, HTTPS is on the wishlist. I just didn't have the time to squeeze it in with v1.0

yegortimoshenko:
FileCast has last been updated three years ago: https://github.com/onyxbits/filecast

Seemingly, safe to close.

alerque:
Agree, this doesn't fit the bill.

Issue #1267, opened 2015-04-03
Add SMSSecure


sedrubal:
Are there any arguments, not to recommend [`SMSSecure`](https://github.com/SMSSecure/SMSSecure)? Otherwise I'll create a pull request.

`SMSSecure` is a forge of [`TextSecure`](https://github.com/WhisperSystems/TextSecure), without `Google API` as dependency (this is why we not recommend `TextSecure`; see #1140 and #552). But without `Google API` there is not problem, isn't it?

Dostoi:
And you can find it on F-Droid : 
https://f-droid.org/repository/browse/?fdfilter=smssecure&fdid=org.smssecure.smssecure

They're planning on making F-Droid their main way of releasing the app

vyp:
I can give a +1 here. Been watching the development of this (was called SecuredText before), and it also recently hit f-droid. The only possible problem I think is that it's a young (as in recent) fork, but go ahead with pull request if you want of course. I don't think it would be a waste time.

strugee:
One issue I see with SMSSecure is that the SMS transport requires a key exchange. This is prone to user error and is also ugly from a UX perspective. Hence why TextSecure got rid of it.

sedrubal:
I see...

And most likely they will have the same problems like TextSecure, that you can't export and revoke your keys and that you can't give your public key through another communication channel and that there will be no difference between writing with a person you validated and writing with a person of whom you have a invalid public key...

But I think it is better to use this app than writing plain SMS...

vyp:
True but as sedrubal says I think that is supposed to be more of an intentional "feature", than a bug.

strugee:
> True but as sedrubal says I think that is supposed to be more of an intentional "feature", than a bug.

(I assume you're replying to me?)

The fact that key exchange is a feature instead of a bug doesn't mean that we can ignore the fact that it's unideal when considering it for inclusion on PRISM Break.

vyp:
> (I assume you're replying to me?)

Yes sorry. But how else can it be done? Is it possible? Because they can't remove support for SMS like textsecure because that's the whole point of the project.

I get your point though, in that it isn't good enough prism break if it has that. But I think it's useful enough to include. While you're right that prism break is all about user friendliness, there's still a wide variety of applications that prism break suggests. And I wouldn't say key exchange is at the bottom of the list in terms of ease of use.

Although maybe I could be wrong/biased. Thoughts?

sedrubal:
:+1:

[...]

PR #1268, opened 2015-04-03
added SMSSecure

sedrubal:
as discussed in #1267. Including english and german translation.


sedrubal:
I'm not sure if you can call `SMS` as a protocol. If so, one should add the protocol `SMS`...

nylira:
Thanks!

PR #1270, opened 2015-04-04
added teamspeak as proprietary

sedrubal:


Zegnat:
I see no problems merging this in, but I have to ask, do we actually offer an alternative for it right now? Do any of the currently listed alternatives work in the same VoIP-chatrooms basis? Putting it in there without offering an alternative will confuse people, I think.

hasufell:
mumble is an alternative.

hasufell:
Ventrilo is also proprietary https://en.wikipedia.org/wiki/Ventrilo and sometimes used in video gaming

sedrubal:
yes, [mumble](https://prism-break.org/en/projects/mumble/) is (nearly) the same as teamspeak in opensource. It is also available for each platform, where teamspeak runs.

Should I also include `Ventrilo` in the same PR? I don't know it, but if it is proprietary, then we should add it...

hasufell:
> in the same PR

probably a separate one

Zegnat:
I don’t think there is a problem if you put it in this same PR. Just make sure they are separate commits.

sedrubal:
voila ;)

wikipedia also mentions [Mohawk Voice](https://en.wikipedia.org/wiki/Mohawk_Voice), but I think, it is quite infamous...

vyp:
Why is being infamous an argument against adding it? If anything, shouldn't it mean we _should_ add it (to the proprietary section)?

Zegnat:
We do not add everything to the proprietary section, and I don’t think that should be necessary either. We would have to list tens of browsers, hundreds of word processors, and thousands of email providers. Including the main players should suffice to get the idea across.

sedrubal:
thank you

vyp:
> We do not add everything to the proprietary section, and I don’t think that should be necessary either. We would have to list tens of browsers, hundreds of word processors, and thousands of email providers. Including the main players should suffice to get the idea across.

Of course, but I think this might be the wrong use of the word infamous? Infamous means it's famous, not 'not famous', so obviously something that is famous should definitely be added to the proprietary section, for obvious reasons. (I have no idea how popular mohawk voice is, hence the question.)

sedrubal:
Oops I'm sorry. I'm not really good in English.

I wanted to say, that it is not really popular so I don't think we have to include it.

Issue #1269, opened 2015-04-03
Add Orweb


sedrubal:
Is there a reason why I can't find [Orweb](https://github.com/guardianproject/Orweb) on prism-break as suggestion under the Android Browsers?

Zegnat:
There was a discussion about Android browsers in #1148, as I personally did not own a smartphone at all I couldn’t comment and was waiting for PRs to come out of that one.

I guess it comes down to wether we want to recommend Orbot + Firefox or Orweb. Other than that I don’t know any reason not to add Orweb.

sedrubal:
Why not recommend both? I'll create a PR...

wifiextender:
Hello sedrubal,

[Orbot](https://gitweb.torproject.org/orbot.git) + Firefox seems rock solid combination as both projects are in constant development, so my vote goes for those two projects.

On the other hand, the default firefox configuration will leak too many fingerprints that will identify you.

Cheers

sedrubal:
but why should we only support one browser?

mastercoms:
Because firefox is the only trusted one.

sedrubal:
I trust [The Guardian Project](https://guardianproject.info/) too. Their slogan is "Mobile Apps and Code You Can Trust" :D

strugee:
Orweb is on PRISM Break now.

Issue #1148, opened 2014-09-10
Android web browsers discussion


anthologist:
I see in the Android section there's only Firefox. The problem is, from v32 they dropped the support for Android 2.2 and for ARMv6 (thanks Mozilla), so personally I'm cut out and probably many other people. Are there any other alternatives?
I was looking at Orweb (https://guardianproject.info/apps/orweb/) based on Shadow Browser (http://www.cl.cam.ac.uk/research/dtg/android/tor/).
Ignoring the security advisory displayed in the orweb page, what do you think  about them?

yegortimoshenko:
Given that ARMv6 build of Firefox on Android has been dropped almost four years ago, there seems to be no point in finding a browser that works on it.

yegortimoshenko:
Closing, because it would be irresponsible of us security-wise to recommend using a device that only runs Android 2.x.

PR #1271, opened 2015-04-05
added Orweb

sedrubal:
as discussed in #1269 and #1148 
Orweb is like the TBB for Android, so I think we should recommend this.


sedrubal:
If this was the reason, why you haven't merged this yet, you can now do it ;)

Issue #1167, opened 2014-10-12
Add Twister to social media


Leodagan:
Twister is an experimental microblogging peer-to-peer platform, implemented as a distributed file sharing system. Being completely decentralized means that no one is able to shut it down, as there is no single point to attack. The system uses end-to-end encryption to safeguard communications. It is based on both BitTorrent and Bitcoin-like protocols and is considered a (distributed) Twitter clone.

http://twister.net.co/
https://en.wikipedia.org/wiki/Twister_%28software%29

PR #1286, opened 2015-04-26
Added twister microblogging platform

P3lUZa:
Added an experimental microblogging platform which uses the bitcoin blockchain, this pull request closes https://github.com/nylira/prism-break/issues/1167 and https://github.com/nylira/prism-break/issues/828

Issue #828, opened 2014-01-13
I just discovered this service: http://twister.net.co/


cryptid11:
It seems a perfect candidate for social network list!!

I read about it here http://www.wired.com/wiredenterprise/2014/01/twister/ but I doesn't know more about it, if anybody has his personal views/further explanation about it I will read it with joy!

yegortimoshenko:
Ultimately, Twister did not succeed and currently sees very little activity (no activity in this year for example): https://github.com/miguelfreitas/twister-core

It's not vibrant, and Mastodon seems to be much more prolific.

cryptid11:
I totally agree ok both of them.

alerque:
This issue closer might have been a bit too aggressive. Obviously it needed resurrecting and reviewing, but a couple months since the last commit and a slow last code year in general don't disqualify a project from inclusion. "No activity this year" is quite misleading in March, and even "no activity in the last year" might not be a make or break issue. I'd like to see something more than this presented as evidence.

Is there any user activity? What do we know about usage? Is it something that still can be used safely (if it ever was)?

Hillside502:
> a couple months since the last commit

The source was last updated 2017-09-09 --- **6 months ago**
https://github.com/miguelfreitas/twister-core/tree/master/src

lukateras:
@alerque I tend to disagree with you in practice: in complex systems such as a blockchain-based Twitter clone, there usually are enough things that warrant improvement/fixes at any given point in time to receive weekly changes. If there's no activity for half a year or so, usually that means the authors are not substantially interested in the project anymore.

I think theoretically your argument is valid, and works for small projects (under 1 KLOC), but we generally don't list very small projects, because these are not usually general enough to be publicly recommended.

I've researched if Twister is substantially used before closing this issue, and all I could find is several comments online proclaiming it's dead, and somewhat ironically, their Twitter account: https://twitter.com/TwisterPtoP

yegortimoshenko:
I also think that burden of proof of activity and maturity lies on the proposal, not on its reviewer.

lukateras:
Well, probably not. I mean, the repo clearly doesn't receive any patches other than to stay compatible with newer dependency versions. 

Twister is 130KLOC of C++ code, which is comparable to a modest browser (e.g. more than all of Servo). With the choice of implementation language, size of codebase, and velocity with which it's currently maintained, there's little doubt that there are remote vulnerabilities that can be used to pwn anyone running a node.

[...]

PR #1287, opened 2015-04-26
Added twister microblogging platform

P3lUZa:
Added an experimental microblogging platform which uses the bitcoin blockchain, this pull request closes https://github.com/nylira/prism-break/issues/1167 https://github.com/nylira/prism-break/issues/828 https://github.com/nylira/prism-break/pull/1286


simonasd:
hum...Some one please check this project out? 
it's becoming popular in China because Twitter is blocked and local SNSs censored seriously.

vyp:
@simonasd on first look, it looks really good. :+1: for me. Especially when people are already using it.

simonasd:
@vyp I'm searching information about this project.its app available in Fdroid.

BTW I just noticed that <a href='https://twitter.com/TwisterPtoP'>@Twister on twitter</a> retweet "China twister user and promoter arrested in Shanghai" days ago. And there is <a href='https://twitter.com/hashtag/FreeLoveChan'>a tag</a> about it. Much to my surprise.

vyp:
@simonasd I saw that too. Don't associate your github account with your actual identity just in case. (Sorry for saying it here, there was no email address associated with your account, and telling you to email me has problems because this is essentially a public forum. And asking for your email here has even more problems.)

simonasd:
@vyp thx very much. I will be much more careful from now on. 

Maybe i can creat a new PGP key and publish it on github, then we can talk safely?

vyp:
@simonasd Yes, but be sure you are creating a new key, instead of just adding an identity to your already existing personal key (if it exists). (Just use your github username for the name of your new identity on the new key?) And use a email address that hopefully cannot be tied back to you (i.e. email address without your name in it). My public key: https://keybase.io/rws/key.asc (at least at time of writing). I'm assuming you know how to do all this?

But that was mainly all I wanted to say anyway. However, it would not hurt creating one anyway if you are willing.

simonasd:
@vyp Ok I got it. 
thank you :)

libres-y-salvajes:
'The assumption is that we should leave the problem of secure
    anonymizing technology implementation to the experts of the field,
    namely the Tor project <https://www.torproject.org/>, and then make
    twister to operate on top of this.
    There is one remaining problem, however, preventing twister from
    operating on top of Tor and thus allowing a more interesting threat
    model where the adversary could be a censoring government. While
    both the registration network (Bitcoin-like) and post
    storage/propagation network (BitTorrent-like) can easily be run on
    top of a Tor proxy, the DHT overlay network uses UDP protocol which
    is not supported by Tor. One possible idea thought is to connect to
    other twister nodes which have full connectivity by TCP and have
    them to relay messages to the DHT network.'


It seems it has to be run on top of Tor in order to be properly
anonymized, and there is still an unresolved issue regarding network
protocols. That is not very obviously stated on the website, though.
Perhaps that might complicate things a bit.
I think configuring protocols and proxies is a little too much if we
want to keep it easy and simple for everyone.

vyp:
@libres-y-salvajes If I'm understanding everything correctly, they don't run on top of tor, precisely because of what you just quoted. It's good that they at least *know* that UDP over tor should not be done, unlike with ZeroNet where they just say they use tor but without *any* explanation how as far as I can see.

> That is not very obviously stated on the website, though. Perhaps that might complicate things a bit.

It's not stated because they don't, therefore it's not a problem.

> I think configuring protocols and proxies is a little too much if we want to keep it easy and simple for everyone.

Looks to me like they were talking about future development (by them), than for *users* to do it themselves.

Provided we want to keep it easy and simple for everyone, by looking at the screenshots, it looks *very* similar to the twitter.com website, which is a *good* thing. For better or for worse, familiarity plays a key role in software being "easier" to people.

The rest of the [faq](http://twister.net.co/?page_id=25) gives the right answers for me. E.g. in the "How secure is twister?" question, they talk about using encryption that was supposedly not compromised by the NSA...

Do understand though that I'm am only advocating for it with the experimental flag, because they themselves give a red beta warning, which should be enough to give the experimental flag to it.

yegortimoshenko:
Blocked by #828.

Issue #1307, opened 2015-05-22
Firefox will start showing ads on new tab page based on browsing history 


vyp:
https://blog.mozilla.org/futurereleases/2015/05/21/help-test-changes-to-new-tab-in-firefox-beta/

---

Random blog post:

http://www.geeksnack.com/2015/05/22/firefox-will-show-ads-based-on-your-browsing-history/

> In Mozilla’s defence, however, we do have plenty of reasons to trust that the nonprofit will keep their promise to not use our data for some nefarious purpose.

Doesn't matter in my opinion. Prism break should not need to trust words.

---

Possibly more information: https://www.reddit.com/r/linux/comments/36txe4/firefox_will_show_ads_based_on_your_browsing/crh4tus Copy pasted:

> Ok, I took a few minutes and reviewed a couple relevant Bugzilla bugs. I'm not a Firefox dev or anything, just someone that needed something to occupy time. From a cursory reading of a couple Bugzilla bugs (which if you follow a the right links from the OP, you will find. I won't link them here because then people will spam them. It took me less than 5 minutes to find them.), I found the following:

> Mozilla is not sending browsing history to the server. What they are doing is getting lists of "Suggested Tiles" along with a list of URLs from the server. If by some algorithm, your history matches that list well enough, they show the tile and notify Mozilla. (I'm not sure about this last part...someone correct me if I'm wrong)

> They prevent Mozilla from passing a list with one item and discovering history by requiring the lists to be long. If I'm reading correctly (I haven't looked at the code), there are a bunch of allowable lists hardcoded into Firefox. The list the server sends with the tiles must be one of those.

> So the only info Mozilla gets is (a) whether your browsing history matches one of the lists of URLs (seemingly hardcoded in Firefox) that shows a tile and (b) whether you clicked on it.

> Do I think this is a good idea? It seems like a step down the wrong path... but at least they considered privacy. That's better than all the other major players, and they still do need to make enough money to exist.

> There's a way to opt out, and it looks like they're automatically opting some people out if they have Do Not Track enabled (even though they point out that they aren't tracking anyone anyway).

But I cannot find the claimed sources for that comment, nor any other official discussion? Probably I just suck at searching.

---

Once some sources or more specific official details are retrieved, the question (obviously) becomes "should Firefox be removed because of this?". Or is this even harmless to privacy or not?

Zegnat:
Duplicate of #1184. Though a lot might have changed since November. This will probably need some restudying again…

[...]

PR #1315, opened 2015-06-03
Tell visitors to use the "Fennec FDroid" F-Droid package for Firefox for

vyp:
Android

Because the "Firefox" package is planned to be dropped soon [1]:

> Note that this package is planned to be dropped from the F-Droid repo
> soon. Please consider using Fennec FDroid instead to be sure of
> continued update support.

[1]: https://f-droid.org/repository/browse/?fdfilter=fennec&fdid=org.mozilla.firefox

---

This was mentioned in #1307 by @mastercoms: https://github.com/nylira/prism-break/issues/1307#issuecomment-104974841, but I did not see a ticket about it.

As copy pasted above (from the Firefox f-droid page), it says to use the Fennec FDroid package instead for continued updates. I've been using it for the past couple of weeks, and have not encountered any (basic) issues with that use. Also, this change already has practical implications, because the Fennec FDroid package [2] has been updated to version `38.0.5`, whereas the Firefox package (on F-Droid) has not and is still on version `38.0`. (And the Firefox play store package is on version `38.0.5`.)

However, the only problem I see is that the Fennec FDroid package page itself says this:

> The app (or some of the builds) might get removed or re-pushed at anytime:
> 
> DO NOT INSTALL unless you know what you are doing!

which I'm not sure about?

Also, I'm assuming this was the correct file to edit? Because I tested with `make test` and it seemed to produce the desired change. I didn't touch the other locales because I saw there were some differences between them and the english version.

[2]: https://f-droid.org/repository/browse/?fdfilter=fennec&fdid=org.mozilla.fennec_fdroid


vyp:
Closing because making a pull request from the default branch (master) (of my fork) complicates things too much. Will make a new pr.

vyp:
#1317.

PR #1316, opened 2015-06-03
Remove Linphone and CSipSimple from the Instant Messaging section

### LINKED ISSUE #1302: no encryption of instant messages in linphone
It seems that chats are not (end-to-end) encrypted in linphone:
https://lists.gnu.org/archive/html/linphone-users/2013-07/msg00021.html
I cannot tell for sure whether this has changed, because I do not use the current version. The website does not indicate that chats are encrypted.

I think this should be noted somewhere. Currently, linphone is described as "Encrypted voice and video __chat__ client for multiple platforms.". Furthermore, it is in the "Instant Messaging" category, not only in the "Video & Voice" category.

vyp:
Closes #1302.

---

Assuming this is the correct change based on discussion in #1302. @real-or-random @juanito003


real-or-random:
Thank you!

Additionally, we should also change the description (see the initial comment in the #1302). Maybe just change "chat client" to "client" there? It would also add a sentence under "Security Notes" like "Note that encryption is only supported for voice and video, but not for instant messaging.".

nylira:
Thanks. Will be pushed live shortly.

vyp:
@nylira I was going to squash, but okay.

Note, I only changed the english ones for the last two commits. Except, I also made a small change to the german file (`source/db/de-projects.json`), which seemed obvious, but I don't know german.

vyp:
@real-or-random Sorry forgot, I'm assuming that (the no encryption for im) only affects Linphone and not CSipSimple? (Because that's what I said.)

real-or-random:
That's what I thought, too. But I've looked it up: csipsimple supports SIP chat, but also only without encryption. So the same applies for csipsimple.

Btw, with the sentence under "Security Notes", I meant rather on the page for the particular tool, not on the overview page. (Like on https://prism-break.org/en/projects/bleachbit/ for instance.) I think this is even more important, even though the hint on the overview page is also good given that several tools are affected (actually all for Android...)

I can also create a pull request by the end of the week. (But I don't mind you doing it, of course. :))

vyp:
Ah I see, the bleachbit example helps thanks. So, because it applies for both csipsimple and linphone, we should just add "security notes" for both of them, correct?

PR #1318, opened 2015-06-03
Note that both CSipSimple and Linphone do not support encryption for IM

vyp:
Ref #1316.

---

Removed the Linphone specific note in the Voice and Video notes section (from 581de2f) in favour of having it directly in the security notes for the program itself (and for CSipSimple too). @real-or-random @juanito003


real-or-random:
What I actually meant in #1316 is to keep also the general note in the video and voice section, because it applies to more than one tool and seems to be a common issue. 
(That's my personal opinion, I'm not the maintainer of this project.)

But nice that you've opened a PR. :)

vyp:
Ah, I removed it from there to avoid duplication and/or confusion. So you're saying it should be on the general video and voice notes section, *and* on the program-specific security notes section as well? (Not rhetorical, I don't see a problem with that, but I'd like to confirm so first.)

real-or-random:
Yes, that's what I'm saying. :)

PR #1320, opened 2015-06-03
Add license url to ProtectedText

### LINKED ISSUE #1010: ProtectedText Licence
Since [version 2.0](https://www.protectedtext.com/site/version2.0) [ProtectedText](https://prism-break.org/en/projects/protectedtext/) now includes licensing information, the JavaScript is [released under MIT](https://www.protectedtext.com/js/main.js).

vyp:
Closes #1010.

---

@Zegnat So just have the same url as the source code, because the license seems to be an embedded comment in the source code anyway?

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -4449,7 +4449,7 @@
   {
     "development_stage": "released",
     "description": "Free online encrypted notepad.",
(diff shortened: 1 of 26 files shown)


nylira:
That works.

PR #1322, opened 2015-06-04
Add DAVdroid

### LINKED ISSUE #885: Add DAVdroid (Android, CalDAV, CardDAV...)
Android does not have support for CalDAV or CardDAV. DAVdroid is an open-source Android application that allows access to CalDAV calendars and CardDAV contact lists in Android through standard Calendar(-based) and Contacts(-based) applications. It is an essential component to connect Android to (for example) an ownCloud or Kolab server, and is therefore software that should be listed on PRISM Break.

Official site:
http://davdroid.bitfire.at/

DAVdroid can be obtained at popular online Android software stores. It is also available for free in F-Droid (linked by the official site):
https://f-droid.org/repository/browse/?fdid=at.bitfire.davdroid
https://f-droid.org/wiki/page/at.bitfire.davdroid

Source code (also linked by the official site):
https://github.com/rfc2822/davdroid

vyp:
Closes #885.

---

Would "Productivity" be the correct category?

Also, judging by the current version number being less than 1.0.0, I gave the experimental tag, would that be a correct reflection? @Zepmann @Cathryne @Raprp @rfc2822 @devvv4ever


nylira:
Productivity is the right category.

rfc2822:
Are there definitions about what "experimental" means?

vyp:
I might be wrong, but in terms of prism break, the experimental flag is given to warn visitors with "newer" or less mature software, because the nature of software prism break commonly deals with is related to security. I think the logic would be that experimental software might be more prone to bugs, so use with caution etc.

In practice atm, if you go to https://prism-break.org/en/categories/android/#productivity, you'll see that DAVdroid is given a yellow "experimental" tag. So the question is, do you think this would be a correct reflection of DAVdroid? Or do you think it is already mature enough for standard/daily use?

devvv4ever:
The experimental tag is fine. Maybe the best choice, since there are still some features that need to be implemented. Nevertheless it's good enough for daily use for ~25k active users ;-) So I don't know what tag would have a better reflection.

vyp:
No worries. If you care, and you want to remove that tag at some point in the future, just give a shout. Also, if I remember, I'll probably make a pull request to remove it once/if you get to version 1.0.0, because that's de facto for "stable".

devvv4ever:
Alright, thanks! Another thing: I would suggest putting DAVdroid to the "Android → File-Storage & Sync" category. It's primarily a sync adapter.

Cathryne:
I have used it for many months without ever noticing problems, and thus consider it `stable` :-)

vyp:
Thanks for the input @Cathryne. I have never used it, so I decided to be cautious by giving it "experimental" when I saw the latest release version number was < 1. However, because #885 is more than a year old, and that it's good enough for daily use by ~25k active users, and with your comment, I agree that it should probably not be "experimental". In my experience, experimental tag is given to much less mature projects, unless someone else wants to weigh in here of course. There will always be "some features that need to be implemented" in a software project (99% of the time), but ~25k active users I think probably means that those features are not the most critical?

So feel free to make a pull request to remove "experimental", referencing this discussion. Otherwise, I will probably do so at some time.

devvv4ever:
I've again looked into the categories and found that "productivity" is pretty fine too. So it's better to leave it where it is ;-)

vyp:
@devvv4ever already been added to File Storage and Sync category, if that is what you meant.

vyp:
Although, it still does not show currently on the website.

Issue #1260, opened 2015-03-23
Broken links in /about/media page


smonff:
Some links are broken because the target website don't exist anymore, or because of editing typos (looks like copy & paste things mostly :+1: ).
I'm a volunteer to fix this. 
Sometimes, backups exists at archive.org : is it ok to use it (I mean, is archive.org a prism-break-compatible infrastructure?) or should I remove the link if nothing else is available?

vyp:
I cannot speak for nylira, but if you want to start doing something, it's highly likely nylira would accept archive.org links.

I'm not sure what nylira would want about the dead links though, so maybe note down which ones they are, but keep them there for now, until he responds.

nylira:
archive.org links would be great.

On Mon, Mar 23, 2015 at 8:21 PM, RWS <notifications@github.com> wrote:

> I cannot speak for nylira, but if you want to start doing something, it's
> highly likely nylira would accept archive.org links.
>
> I'm not sure what nylira would want about the dead links though, so maybe
> note down which ones they are, but keep them there for now, until he
> responds.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nylira/prism-break/issues/1260#issuecomment-84970238>.
>



-- 
Peng Zhong
Interaction Designer & Developer

Designer at http://nylira.com
Curator at https://prism-break.org
GitHub: https://github.com/nylira
LinkedIn: https://www.linkedin.com/in/nylira

strugee:
@vyp looks like this is resolved given the above PR, yes?

smonff:
[Another 404](http://www.srf.ch/k.text-blocktur/im-fokus/weblese/surfen-ohne-dass-die-nsa-zuschaut). Should link to http://www.srf.ch/kultur/netzwelt/surfen-ohne-dass-die-nsa-zuschaut.

Ref. #1324

vyp:
There are still some broken links on the page.. see #1324.

smonff:
[Schieb, Jörg.“Wie bewege ich mich sicher im Netz?](http://www1.wdr.de/fernsehen/aks/rubriken/angeklickt/angeklicktsichersurfen100.html) can link to https://web.archive.org/web/20141003211621/http://www1.wdr.de/fernsehen/aks/rubriken/angeklickt/angeklicktsichererimnetz100.html

smonff:
[...]

PR #1324, opened 2015-06-05
Fix some broken media links with archive.org backups

vyp:
Ref #1260.

---

I wouldn't say it "closes" #1260 because there are still these broken links which do not seem to have archive.org backups (they either were not indexed or the snapshots themselves were all broken too):

- Röll, Nadja. http://www.srf.ch/k.text-blocktur/im-fokus/weblese/surfen-ohne-dass-die-nsa-zuschaut
- Luckerson, Victor. http://business.time.com/2013/06/20/the-anonymous-internet-privacy-tools-grow-in-pop.text-blockarity-following-nsa-revelations/
- Informare per resistere. http://www.informarexresistere.fr/2013/07/08/blocca-il-controllo-del-governo-usa-s.text-blockle-tue-attivita-online-liberati-da-prism/

Additionally, this one seems to be behind a paywall (and same with the archive.org versions), which I'm not sure is a problem but might be:

- de Sá, Nelson. http://www1.folha.uol.com.br/colunas/nelsondesa/2013/07/1311846-para-se-livrar-do-google.shtml

But what can you do other than remove them right? So I'll let you decide if this completely solves #1260 or not.

### DIFF
--- a/source/views/about/media.jade
+++ b/source/views/about/media.jade
@@ -227,7 +227,7 @@ block content
       span.date 2013-07-05.
     div
       span.author Neagle, Colin.
-      a.title(href='https://www.networkworld.com/community/blog/how-opt-out-prism-nsas-spying-program')
+      a.title(href='https://web.archive.org/web/20140113142442/http://www.networkworld.com/community/blog/how-opt-out-prism-nsas-spying-program')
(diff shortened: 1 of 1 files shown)


nylira:
It doesn't entirely solve it, but it doesn't all have to be figured out at once.

PR #1326, opened 2015-06-05
Add DAVdroid to File Storage & Sync category

vyp:
Ref #1322.

---

Not sure if it's a traditional "File Storage & Sync" app, but pr on request of @devvv4ever in #1322. Your call in the end, but do note that @devvv4ever is a DAVdroid developer.

@rfc2822 (also DAVdroid developer)


nylira:
Sounds good.

PR #1331, opened 2015-06-07
Sort db json files by 'slug' key

vyp:
It clearly looks like everything in the `source/db/*.json` files are supposed to be alphabetically arranged, but probably overtime there's been a lot of mix ups. Obviously, strictly keeping it alphabetical all the time is probably a waste of time. But I think I and maybe others would find it easier for the future if it is a bit more sorted. (Even just to navigate, because they are large files in terms of number of lines.)

Ftr, I used a small script to do this and hence why you can see some tabs have been changed to spaces (as in, just the whitespace indentation, not actual content). Only changed `source/db/*.json` files because I don't think any other files are large enough to consider this anyway.

Tested with `make reset`, doesn't seem to mess anything up.

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -51,6 +51,29 @@
     "slug": "adium"
   },
   {
+    "development_stage": "experimental",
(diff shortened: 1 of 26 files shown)


vyp:
Also btw, check the diff in the terminal or on a local client or something because it might be too big to view in the browser (it crashes mine).

nylira:
Sure, that works.

PR #1340, opened 2015-06-14
Remove experimental tag from DAVdroid

vyp:
Ref #1322.


Zegnat:
Merging this, mostly because I don’t believe there is any weird encryption stuff going on here. Experimental is more aimed at encryption tools that might fall through completely at some point or are non-mature implementations. I don’t think that applies here (?).

vyp:
Yes I think you are correct.

PR #1347, opened 2015-06-19
Replace LibreWRT with libreCMC

### LINKED ISSUE #1346: Add LibreCMC
LibreCMC supports more hardware than LibreWRT, and it is FSF-approved.

vyp:
Because they "have joined forces" and "future work will be done under
the libreCMC name" [1].

libreCMC is also FSF-endorsed [2].

Closes #1346.

[1]: http://librewrt.org/index.php?title=Main_Page
[2]: http://www.fsf.org/news/fsf-adds-librecmc-to-endorsed-distros-list


nylira:
Thanks!

PR #1359, opened 2015-07-16
clarify that uBlock Origin is not a fork of Adblock Plus

### LINKED ISSUE #1358: Possibly misleading statement under 'Web Browser Addons' section
Under the ['Web Browser Addons' section] (https://prism-break.org/en/all/#web-browser-addons), the final paragraph under 'Notes' heading reads: 
> Why is Adblock Plus not recommended? Adblock Plus shows “acceptable ads” by default, which works against the purpose of the add-on. Either disable acceptable ads or use the uBlock fork instead.

This last statement seems to imply that uBlock is a fork of Adblock Plus, which it is not. (This is repeated in a number of translations also.)

vyp:
Closes #1358.


nylira:
Thanks! Building and will be live soon.

PR #1371, opened 2015-07-27
Update fr-projects.json

Fend4self:
adblock-edge + adium  FR translaction correction+addition

### DIFF
--- a/source/db/fr-projects.json
+++ b/source/db/fr-projects.json
@@ -16,7 +16,7 @@
       {
         "name": "Android",
         "subcategories": [
-          "web browser addons"
+          "Web browser addons"
         ]
       }
     ],
@@ -43,7 +43,7 @@
       {
         "name": "OS X",
         "subcategories": [
-          "instant messaging",
+          "Instant messaging",
           "IRC"
         ]
       }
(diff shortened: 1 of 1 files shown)


vyp:
Pretty sure the translations to "web browser addons", "experimental" and "instant messaging" are not needed because they are already done in [source/locales/fr.json](https://github.com/nylira/prism-break/blob/master/source/locales/fr.json), but I may be wrong.

nylira:
What @vyp said is right. Please edit out those changes, and then I'll sync this.

Fend4self:
Sorry, it wasn't clear to me wether we should edit from ./source/db/*-projects.json or ./source/locales/*.json

source/db/fr-projects.json
```diff
@@ -43,7 +43,7 @@
       {
         "name": "OS X",
         "subcategories": [
-          "Instant Messaging",
+          "instant messaging",
```
nylira: Capitalize this to revert it to the previous version.

source/db/fr-projects.json
```diff
@@ -16,7 +16,7 @@
       {
         "name": "Android",
         "subcategories": [
-          "Web Browser Addons"
+          "web browser addons"
```
nylira: And this one too!

Fend4self:
Done

PR #1376, opened 2015-07-28
 synchronize missing/unequal metadata across all locales

vyp:
After this change, the only differences (I think) left are:

1. "fr-projects.json" has Wikipedia urls for the "Disconnect",
   "EtherCalc" and "Kamailio" projects, which no other locale has.

2. "zh-CN-projects.json" has translations for the "I2P" and "Ixquick"
   projects' "notes" section, which no other locale has, as introduced
   by b5495f7 from #932.

  - Notice that in the diff of b5495f7, the "notes" section for both
    these projects were empty before the change.

3. "fi-projects.json" has a translation for the "Off-the-Record
   Messaging" project's "notes" section, which no other locale has, as
   introduced by 1bd7801 from #1177.

  - Notice that in the diff of 1bd7801, the "notes" section for this
    project was empty before the change.

1 however should not be a problem because it does not look like there
are Wikipedia articles for these projects under any other language,
including English.

Note that this change does not actually check for *outdated*
translations (obviously). This change has only checked if a string type
key is empty when it should not be, or if an array type key (i.e.
"protocols" and "categories") is not equal to the rest of the keys from
all the other locales.

---

So I noticed that there were some missing or different metadata across all the `source/db/*.json` files. The reason there are two commits is because, whilst doing this, I noticed that the indentation for Syncthing was off (which was also my fault). So you can view both commit diffs separately so that the syncthing indentation stuff does not clutter the actual synchronizing diff.

I'm not sure what to do about 2 and 3 though, perhaps google translate to english and insert the results into all the other languages? Pinging @sedrubal (introduced 2) and @maqp (introduced 3) for english translations of those changes if possible. But I think it's better if they're just left alone, because the chinese translations at least seem to be china-specific. What do you think?

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -144,7 +144,7 @@
     "name": "ArkOS",
     "tos_url": "",
     "url": "https://arkos.io/",
-    "wikipedia_url": "",
+    "wikipedia_url": "https://en.wikipedia.org/wiki/ArkOS",
     "protocols": [],
     "categories": [
       {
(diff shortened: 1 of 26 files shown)


vyp:
Sorry @sedrubal, not you, @simonasd introduced 2.

sedrubal:
NP ;)

nylira:
Thanks for working on this. I think we can leave 2 and 3 untranslated for now.

PR #1388, opened 2015-08-26
Added AntiPrism project reference

antiprismca:
Hello, we would like to suggest adding our project to prism-break.org.


Zegnat:
I am wondering if this really should be filed as a replacing / live OS. Reading the web page it seems to be specifically made to be run on a separate machine or home server? Should we list AntiPrism only as a server OS, perhaps?

antiprismca:
Yes, this is "media server" (maybe add a category?) enhanced with anonymization functionality, it can act as an anonymizing proxy or an onion router, plus we can host hidden websites and anonymous torrents, so it is really more like a server. And, we support Raspberry Pi / Pi2. 

Will you modify the files, or we re-commit them under another category?

Thanks!

vyp:
I agree with it being in the "Servers" category with subcategories of "Raspberry Pi" and "Operating Systems", but it does not seem to be a live system by reading the website, am I wrong? So in that case, shouldn't it only be in the "Servers" category, instead of also being under "Windows", "OS X" and "GNU/Linux"?

If you agree with that, then I'd be happy to make the necessary edits if you don't have the time to do so. But because it looks like it's not possible to make a pull request on a "fork" (a repository which is considered a "fork" by github), this PR would first have to be accepted, and then the correction will need to be made. @Zegnat that should not be a problem though because the site is rebuilt manually by nylira everytime there is a change, so there shouldn't be any chance of visitors seeing the wrong information even temporarily.

antiprismca:
You are right, it suits more as a "Server" with "Raspberry Pi" and "Operating Systems" categories. Actually, we would keep "GNU/Linux" as it fits it too. As about "Live",  "Windows" and "OS X", we thought originally to put it this way because is starts like a live OS (similar to Tails) from a USB key or a memory card configurable in Linux, Windows or OS X, but we suggest installing it from the media to HDD/SSD anyways.  So, if it is OK with you, please correct the files to reflect the changes you proposed.

Thanks!

vyp:
> because is starts like a live OS (similar to Tails)

So I was wrong, sorry. But if I understand correctly, you do not want it under "Operating Systems (Live)" because you suggest intalling it to the hard drive anyway?

My objection with listing it under the "Operating Systems" subcategory under the "Windows", "OS X" and "GNU/Linux" categories, is because it's purpose is specialised (media server), rather than a general purpose OS. If you look at https://prism-break.org/en/all/#operating-systems, you can see that systems like "ArkOS", "CentOS", "FreedomBox" etc. are only in the "Servers" category. I think it fits in with them here, and that listing it under the "Windows", "OS X" and "GNU/Linux" categories alongside general purpose systems such as Arch, Debian and Fedora would be inconsistent.

If you agree with that, I think @Zegnat also agrees with that too?

So the next thing would be to accept antiprism, then I can make another pull request with the correct categories if needed. @Zegnat I think AntiPrism serves a unique enough purpose (a server operating system dedicated solely to media serving) that it should be listed on the site. The information on the homepage looks quite good to me. The only thing that stood out was the use of truecrypt, but it's only optional, so it shouldn't stop antiprism from being listed (perhaps a mention under "Notes"?).

Zegnat:
Yes, @vyp, that is almost exactly what I was thinking. If someone is looking for a Mac OS X or Windows alternative, they are most probably looking for an OS they can use for day-to-day tasks: an office suite, calendars, maybe games. This just doesn’t seem like the kind of OS AntiPrism is aiming to be.

As for the live OS: I was imagining the kind of thumb drive you would have with you incase you needed access to encrypted emails or are travelling in countries with restricted internet access. Again, a USB drive with AntiPrism doesn’t seem to be the thing I am likely to look for in this case. Or at least cannot compete with Tails and JonDo.

But for the Servers category it is excellent, much like (as you said) FreedomBox.

> <i>[…]</i> this PR would first have to be accepted, and then the correction will need to be made.

I am currently travelling out of the country and don’t have my computer with me to make any changes here. If this pull gets updated I can use GitHub’s auto-merge, otherwise it will have to wait for me to get back home or @nylira to do this.

vyp:
@Zegnat I have time right now (and tomorrow), it's only a few edits, will take like two minutes. Basically what I'm saying is, this PR can be accepted, and @nylira just has to wait until the category corrections are made and accepted before rebuilding the site, so visitors will not see incorrect information, even if only temporarily. Ideally @antiprismca should make the change, but I'm not sure he/she wants to, or has the time, or something?

The reason this needs to accepted first though is because it does not seem to be possible to make a PR on @antiprismca's repo, because it's a "fork", not an original repo. The other alternative would be if I make another entire new correct pull request, but that's a few minutes more work, and I'm feeling lazy atm.

nylira:
@vyp let's go with the former. Everything else seems to be in order.

PR #1390, opened 2015-08-29
AntiPrism should only be listed under the "Servers" category

vyp:
Ref #1388.

---

@antiprismca So I only put it under the "Servers" category as @Zegnat agreed with me in #1388. I didn't  change anything else.


nylira:
Thanks!

Zegnat:
Nice work @vyp, thanks for jumping in @nylira!

PR #1393, opened 2015-09-04
add some "notes" for zh-CN and en

simonasd:
Update zh-CN-projects.json
1. fix "notes" section of bookie and Ixquick
2. add "notes" for Tor. This section is china-specific, same as I2P.

Update en-projects.json
1. add notes for Ixquick and Startpage

PR #1394, opened 2015-09-04
Synchronize added metadata from b32c856 (#1393) for Ixquick and

vyp:
Startpage

This commit also changes "It also provides [...]" in the "notes" section
for Startpage to "Ixquick also provides [...]" so that there's less
confusion as to which project provides the service, despite the name of
the service giving it away anyway.

Ref #1376.

---

Thanks to @simonasd for these updated and helpful notes of course.


nylira:
Thanks!

Issue #1313, opened 2015-06-02
SourceForge tempering FOSS with adwares and more


ghost:
Not sure if this is completely off-topic but since some of the software that PB recommends is hosted on source forge, maybe recommend the user to try to get the binaries directly from the project's website.

> http://tech.slashdot.org/story/15/06/01/1241231/sourceforge-and-gimp?utm_source=rss1.0mainlinkanon&utm_medium=feed
> 
> In response to a Google+ post from the Gimp project claiming that "[Sourceforge] is now distributing an ads-enabled installer of GIMP," Sourceforge had this response: "In cases where a project is no longer actively being maintained, SourceForge has in some cases established a mirror of releases that are hosted elsewhere. This was done for GIMP-Win.
> 
> Submitter's note: Gimp is actively being maintained and the definition of "mirror" is quite misleading here as a modified binary is no longer a verbatim copy. Download statistics for Gimp on Windows show SourceForge as offering over 1,000 downloads per day of the Gimp software.
> 
> In an official response to this incident, the official Gimp project team reminds users to use official download methods. Slashdotters may remember the last time news like this surfaced (2013) when the Gimp team decided to move downloads from SourceForge to their own FTP service. "Therefore, we remind you again that GIMP only provides builds for Windows via its official Downloads page." Note: SourceForge and Slashdot share a corporate parent.

Zegnat:
I guess this means not only do we need to double check Google Code links but also go through all SourceForge links…

mastercoms:
But they reverted their decision.
https://sourceforge.net/blog/third-party-offers-will-be-presented-with-opt-in-projects-only/

They will not put adware into software that does not opt in to it.

Raprp:
For me only takes one stunt like this to stop trusting a project. I believe that for the right money they would temper the software again and be very cautions to avoid getting caught (tracking instead of ads for example).

But that's just me.

mastercoms:
Yeah, I have the same opinion towards the issue, but I wanted to get new information out there.

vyp:
That's not reverting a decision, that's only damage control because it became so public. They would have never have posted that if GIMP did not make a public post about it. Raprp, what are some projects that are hosted on sourceforge (which are recommended by pb)?

Raprp:
With a quick search on the source I found:

OpenPGP - http://sourceforge.net/p/enigmail/source/ci/master/tree/genxpi
IndiMail - http://sourceforge.net/p/indimail/code/ci/master/tree/
[...]

PR #1395, opened 2015-09-04
Update some SourceForge links

vyp:
Updated projects include:

- Mumble.
- Retroshare.
- Seeks.

Ref #1313.

---

Judging from https://github.com/nylira/prism-break/issues/1313#issuecomment-108727145, Enigmail, IndiMail and PasswdSafe still use SourceForge for source code hosting, although there could be more, I haven't checked. (Note that Bleachbit has indeed switched to Github, just the site is still hosted on sourceforge.)

Also, the license url for enigmail ([project link](https://prism-break.org/en/projects/enigmail/)) is dead for me, what about you? (I couldn't find an updated license look for it with a basic search, except http://sourceforge.net/projects/enigmail/ does say that the license is GPLv2 and MPL.)


nylira:
There's a license page for Enigmail outside of Sourceforge here: https://www.enigmail.net/home/license.php

Issue #1255, opened 2015-03-20
code of conduct


hasufell:
Since the prism-break community is already quite large, I'd say we should set up a small code of conduct and add it to the README, so that people are aware of what is expected of them when discussing on the issue tracker.

Examples include:
https://en.opensuse.org/openSUSE:Conference_code_of_conduct
https://wiki.gentoo.org/wiki/Project:Council/Code_of_conduct
https://www.debian.org/code_of_conduct

It probably shouldn't be that verbose, but short and expressive.

I am thinking of something along these lines:
* be helpful and collaborative
* respectfully disagree
* don't get personal, stick to the factual level
* don't unnecessarily incite drama or negativity

In my opinion strong and heated arguments are helpful, if they stay on the factual level and help moving the discussion forward. That's why I wouldn't add unnecessary "love each other" phrases (although people may do that if they want).

Draxler:
:+1:

strugee:
> That's why I wouldn't add unnecessary "love each other" phrases (although people may do that if they want).

IMHO, we should at least recommend this, even if we don't require it. Some people have a lower tolerance for language directed "against" them, and it'd be nice to try to help them, even if we don't bend over backwards.

PR #1400, opened 2015-09-16
Adopt a Code of Conduct for project participants

alerque:
PRISM-Break is largely driven by community participation in discussions
surrounding the inclusion or exclusion of various solutions. Given that
user privacy is (by definition) a very personal matter and that the
security community in general tends to have strong opinions, my
observation is that many issues that come up are fairly highly changed
even before somebody looses their cool or makes a nasty remark.

I think it would behoove the project to officially adopt a Code of
Conduct document that can be considered binding on participants in
discussion. Getting this into the repository early before any particular
major issue between participants comes up gives everybody something to
reference and a relative standard by with to judge a situation. This
could potentially save some grief vs. the possibility of working out the
details of such a code in the middle of a disputed discussion.

The Open Code of Conduct seems like a likely candidate. I'm not a huge
fan of some of the items as worded but for this project I think it will
be fine. See http://todogroup.org/opencodeofconduct/ for details.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -130,3 +130,8 @@ If you wish to mirror this site, [nylira/prism-break-static](https://github.com/
 
  - [https://www.sedrubal.de/service/prism-break/en/](https://www.sedrubal.de/service/prism-break/en/)
  (credit: [sedrubal](https://github.com/sedrubal))
+
+# Code of Conduct
+
+This project adheres to the [Open Code of Conduct][code-of-conduct]. By participating, you are expected to honor this code.
+[code-of-conduct]: http://todogroup.org/opencodeofconduct/#PRISM-Break/abuse@prism-break.org


hasufell:
#1255

hasufell:
I think no one is going to read that wall of text.. we need a few concise and expressive bullet points on the front page.

alerque:
@hasufell I agree that not everybody will read this up front, but then again the thing that is in the CONTRIBUTING.md document per this suggestion is just a one liner saying that there _IS_ a code of conduct. I believe this is the key feature. Most people aren't going to even need to read this because they behave decently enough anyway, but just the one liner in the file noting that there is such a thing should put those who do see it on notice that they should behave professionally and tip any would be cranks off that this project might not be the best place to act up. What specifically the points are is less important at that stage.

The other use it serves is when a specific issue does come up, and this is where having the somewhat more verbose edition around should prove to be a good thing. If something does happen there won't be a need to argue about what is or isn't appropriate, there is already a basic outline is and you can point out the relevant point and move on. This saves accusations of making up the rules as you go along or them being a matter of interpretation (they still are of course, but much less so than if it was a more general and less detailed outline).

strugee:
+1, this is long overdue. <3

Issue #1299, opened 2015-05-16
RequestPolicy is incompaible with Firefox 38.0.1


kristinegood:
First of all, thank you for maintaining prism-break. I have been slowing replacing proprietary solutions I've used for years with better, privacy-centered, open-sourced one. 

I just updated my Firefox browser and RquestPolicy will not work now. Since it's no longer maintained, is it time to find an alternative?

I found these so far:

[uMatrix](https://github.com/gorhill/uMatrix)

[policeman](https://github.com/futpib/policeman)

[RequestPolicyContinued](https://github.com/RequestPolicyContinued/requestpolicy/)

passiflora:
uMatrix's GitHub page says "Definitely for advanced users." right from start...

from github's graphs RequestPolicyContinued appears to be much more active than policeman

Zegnat:
RequestPolicyContinued was planned to replace it, see #1153. The problem back then was that it did not actually have all functions implemented. You would have taken a step back in functionality quite a lot. Not sure what the state of it is now, but if the old RequestPolicy really no longer works we may not have a choice.

nodiscc:
Hi @Zegnat I help maintaining the [RequestPolicyContinued Github project](https://github.com/RequestPolicyContinued/requestpolicy/) (though @myrdd is doing all the heavy work), so feel free to ask any question.

The current most important problem with RequestPolicyContinued is that it's unable to import custom rule sets from legacy 0.5x versions, and users may not want to lose their rules during the transition. Support for importing old rules will likely be [part of the next release](https://github.com/RequestPolicyContinued/requestpolicy/issues/354#issuecomment-103240962).

The old [context menu entry is gone](https://github.com/RequestPolicyContinued/requestpolicy/issues/353) and has been replaced with a toolbar button/menu, which is an important UI change. It provides the same functionality as the legacy 0.5 menu, except:

There is no support for [strictness](https://github.com/RequestPolicyContinued/requestpolicy/issues/474) in the menu, although you can decide in the prefs if you want domain-level or subdomain-level strictness in the menu actions. This is a global setting though.

The addon has it's share of [bugs](https://github.com/RequestPolicyContinued/requestpolicy/labels/bug), some of which are common with the 0.5x version. Moving the codebase to the 1.x branch (beta version originally available on requestpolicy.com) allowed to get rid of some [legacy version bugs](https://github.com/RequestPolicyContinued/requestpolicy/issues?q=label%3A0.5.x+is%3Aclosed)

[...]

PR #1403, opened 2015-09-20
Add uMatrix (Firefox addon)

geplus:
https://addons.mozilla.org/it/firefox/addon/umatrix/?src=dp-dl-othersby

"Point & click to forbid/allow any class of requests made by your browser. Use it to block scripts, iframes, ads, facebook, etc."

 GNU General Public License, versione 3.0

I think it shares part of its code with uBlock


vyp:
Hmm, the diff is huge... I'm not sure but it says you want to merge into the 0.1.0 branch instead of what I think should be the master branch. Can you confirm?

Zegnat:
I seem to recall earlier discussions on this where it was decided that uMatrix was such a power-user tool that we were more inclined to recommend RequestPolicy instead. Can’t find any links right now though. Thoughts?

vyp:
@Zegnat You are correct, discussion was in #1299. I agree with gorhill that uMatrix is for advanced users, but perhaps we could still recommend and give it a warning in the notes section? (btw, requestpolicy is already recommended on the site, if you meant one or the other.)

vyp:
@Zegnat Also, this unrelated, but this reminds me, we probably want to update requestpolicy to https://github.com/RequestPolicyContinued/requestpolicy/ as @nodiscc said. What do you say?

Zegnat:
800a30d mentions uMatrix as a more advanced alternative to Request Policy. I think I am going with that for now.

geplus:
Can you mention it as an alternative to request policy in the Web Browsers Addons section?

Zegnat:
@geplus, that is exactly what I did in 800a30d. You can see I added the following to `web-browser-addons-notes`:

```
<p>A more advanced alternative to <strong>Request Policy</strong> would be <a href='https://github.com/gorhill/uMatrix'><strong>uMatrix</strong></a>.</p>
```

The change will go live at the next update.

geplus:
OK, thank you!

PR #1409, opened 2015-09-27
Update Privacy Badger status

anthologist:
It's now out of beta: http://www.pcworld.com/article/2961068/privacy/eff-tracker-smashing-privacy-badger-exits-beta.html

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -4833,7 +4833,7 @@
     "slug": "postfix"
   },
   {
-    "development_stage": "experimental",
+    "development_stage": "released",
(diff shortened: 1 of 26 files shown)


nylira:
Thanks!

alerque:
Umm...next time could we get this sort of thing merged into a single commit per action rather than per file affected? This makes the history hard to follow and operations such as reverts or git blame which allows you to connect changes that happened together nearly impossible. @nylira I would even be inclined to rebase and squash this in master since it's still fresh and the last thing that happened in the puiblic repo, but that's up to you. Once people start building on the lastest master that starts getting disruptive.

PR #1425, opened 2015-10-11
Translate the introduction of SMSSecure

simonasd:
Translate the introduction of SMSSecure which was introduced by [#1268](https://github.com/nylira/prism-break/pull/1268)


Zegnat:
Thank you very much, translations are always more than welcome!

simonasd:
My pleasure :)

Issue #1344, opened 2015-06-16
Recommendation free social networking engine „Elgg“


Pasley70:
Elgg is an award-winning social networking engine, delivering the building blocks that enable businesses, schools, universities and associations to create their own fully-featured social networks and applications.

https://www.elgg.org

Thank you to add to selection Social Networks

vyp:
@Pasley70 was Elgg previously known as "Lorea"?

Pasley70:
@vyp "Lorea" was a project, not a software. See https://en.wikipedia.org/wiki/Elgg_(software)

vyp:
@Pasley70 Thanks. I asked because Lorea was actually recommended on Prism Break before. In fact, Lorea was only taken down a few days ago, see #1070. I was trying to find out more about it but could not find much. I remember I somehow ended up at Elgg, but I could not find a link between Elgg and Lorea and concluded that they were unrelated. Yes, my "research" was not that comprehensive.

Elgg describes itself as a "social networking engine". However, I'm assuming the prism break project maintainers would want to list something that their visitors can use/sign up for, or could self-host, or even potentially use in a p2p manner. I'm not sure what benefit there will be by *directly* listing the software that provides "the core components you need to build out socially aware applications".

So do you know of another "Lorea" that is still active today? My logic is that such a network, if it exists, will have a chance at getting onto prism break because it also uses Elgg like Lorea did.

Pasley70:
I think you can compare Elgg with friendica. Perhaps listing Elgg on this platform ist interesting for people who would host their own special-interest-communities for a specific group of people.

strugee:
@Pasley70 I disagree. Friendica is a complete product that you can actually run on your own server. Elgg, AFAICT, is not.

Pasley70:
@strugee https://www.elgg.org/requirements.php

strugee:
@Pasley70 OK, I may have misunderstood their website, which made it sound like a bunch of libraries.

I should mention that the page you linked to did not lead me to that conclusion. Libraries have requirements too, you know. In the future, it would be helpful if you could link to something more substantive. Thanks!

@vyp from http://learn.elgg.org/en/stable/index.html, it seems like Elgg actually can be installed as a full application. However, seems like it's only used for "private social networks", e.g. internal to some organization. [Support for federation][1] seemed to be weak to begin with and has since stalled out. Given that, I don't see why we should list Elgg when we have networks on the list that actually federate.

[...]

PR #1431, opened 2015-10-19
add protectednet.io

evgenyigumnov:
Please add my web service


hasufell:
sounds like this service should definitely by [AGPL](https://en.wikipedia.org/wiki/Affero_General_Public_License) (preferably version 3 since version 1 is weirdly incompatible with GPL-2), not GPL
see
```
Both versions of the AGPL were designed to close a perceived application service provider "loophole" (the "ASP loophole") in the ordinary GPL, where, by using but not distributing the software, the copyleft provisions are not triggered.
```

evgenyigumnov:
Sorry. I am not a specialists in licenses ;)

Draxler:
@evgenyigumnov 

Could you corected URL in your title. 

> It's : protectednet.io    

> Not: portectednet.io  as in your tilte

Regards

evgenyigumnov:
misspell )

Draxler:
@evgenyigumnov 

Yes i know, but people need to actual go there and see what you offer :smile: 
That's why i suggest to corect misspel where need it.

evgenyigumnov:
Is it possible to correct comment for pull request???

evgenyigumnov:
WOW ! It is possible. I was fix it!
Thank you.

Probably anybody could translate my article about ProtectedNet from Russian to English?
My English level so small -(

This is my article in Russian language: http://evgeny.igumnov.com/wiki/2/638116

Draxler:
:+1:  :)

Draxler:
is this any good? This was translated with one realy good translater. Nothing was edited.
i begin at the start:

In this article I want to tell you how I came up with the idea of ​​web-service secure social networks. How to implement it and on what technologies. I will share technological solutions to problems during the development of the service.

The idea of ​​this project i had after reading the current state of affairs in the field of encryption. I ran into two problems.

On one hand, this is a complete lack of protection of users from the leakage of information from social networking sites, instant messengers and more. For example Skype and Telegram keeps all correspondence in their own servers and at the request of the Government to provide any data from these correspondences. Still remember, the recent hacking site for adultery Ashley Madison, which sailed user data.


On the other hand we all know that the higher the level of protection, the less pleasant to use such a product. For example, to sign and encrypt email using PGP, require special software and skills.
Using Telegram, you are obliged to establish their messenger and shine your sim card and also a secret chat mode, with only one user group mode is not. It is also removable, you will not see what you chatted earlier. ........

evgenyigumnov:
Sounds good

evgenyigumnov:
Is it https://translate.google.ru ? )

Draxler:
This is one of the best translating tools i could foud. it's for Firefox or its forks.

https://addons.mozilla.org/en-US/firefox/addon/fast-translation/?src=userprofile

evgenyigumnov:
I think we need in human translation.

Draxler:
yes i know,  it's just a personal help tool.. :smile:

hasufell:
> Sorry. I am not a specialists in licenses ;)

It basically means that under your current license someone could run a service based on your code without actually making the (modified) source code public. Because just running a service does not mean to _distribute_ the code. AGPL fixes this loophole.

evgenyigumnov:
I dont care about this. If you would like you could download codes and buy server for run my scripts.
If you dont wont this boring and expencive action then you could simple use my web service :)

hasufell:
> I dont care about this. If you would like you could download codes and buy server for run my scripts.
> If you dont wont this boring and expensive action then you could simple use my web service :)

It allows people to fake your service while they are in fact running proprietary software behind it.

evgenyigumnov:
If some body would like do it then AGPL dont stop them ;)

hasufell:
> If some body would like do it then AGPL dont stop them ;)

Following that argumentation you could as well say GPL-2 (which you are using) doesn't stop anyone to honour copyleft either.

evgenyigumnov:
Unfortunately, yes, doesn't stop.
I am from Russia and I dont trust and dont belive in people ;)

hasufell:
> I am from Russia and I dont trust and dont belive in people ;)

hahaha :D 

My whole point is... if you already use a GPL-type license, just do it right. This would also theoretically allow people to sue a**holes better... and that doesn't really have to be you, but basically anyone who uses your software. It benefits others.

Zegnat:
The current PR seems to set it up under Social Networks, but I don’t think it goes there as it doesn’t replace any of the proprietary platforms in that category in any way. This should probably go under media publishing, but even then I am not sure.

This looks more like a replacement for MEGA, MediaFire, etc. None of those are listed on PRISM Break, and I am not sure what the usecase is for putting them on here. Thoughts?

evgenyigumnov:
You could create private, small and secure social network. It is not open network... It is not global social network for all users in internet. You could use this service for your small team for example.

strugee:
@evgenyigumnov in that case, what advantages does this software provide over the existing list?

Seems like it's useful in a different context, but not helpful as a replacement for large-scale proprietary social networks. And, the non-strong copyleft makes it a little bit unideal.

IIRC, this was one of the main concerns when we rejected Elgg (#1344).

evgenyigumnov:
@strugee Look at https://www.protectedtext.com I am using the same idea only for social network communications.

strugee:
@evgenyigumnov you're missing the point I'm making. PRISM Break is trying to make nation-state surveillance uneconomical via the mass adoption of strong cryptography and decentralization. If, as you say, it's not a "global social network," how does that fit in with PRISM Break's stated goals?

Note that I'm not saying that your software isn't good or isn't useful. I'm just saying that I don't see how it fits into PRISM Break.

evgenyigumnov:
I was change my mind ;)

I was setup AGPL license for source codes.
I was make registration and service use for free on web-service https://protectednet.io

hasufell:
How do you deploy this service on your own server? I don't see any instructions.

Zegnat:
I am going to close this. There are no instructions for anyone to run this themselves, while some set-up instructions are definitely required (e.g. setting the Application Secret for the Play framework). I am not even sure how the hosted version works… I created `https://protectednet.io/prism-break` and a “network post” I made isn’t even visible on my second test account. Maybe it is a bug, but that clearly shows there is a lot of work and testing still to be done.

I do not have Java installed, so I can’t even try setting it up, but I don’t think we can recommend it as a self-hosted social network for friends at this stage.

PR #1441, opened 2015-11-26
remove "en-US" part from 'addons.mozilla.org' links

### LINKED ISSUE #1440: Fix AMO Links
It would be good if you could change the addons.mozilla.org links to remove the localization part. For example:
https://addons.mozilla.org/en-US/firefox/addon/adblock-edge/
to
https://addons.mozilla.org/firefox/addon/adblock-edge/
If you visit the section "Web Browser Addons" in another language (e.g. https://prism-break.org/de/all/#web-browser-addons), the links still point to the american/english version of the site. Removing the "en-US" part leaves it to AMO to detect the best language and redirect there (e.g. https://addons.mozilla.org/de/firefox/addon/adblock-edge/ for me).
It's a very minor issue but it just makes sense. Thanks for the great work.

vyp:
So that AMO automatically detects an appropriate language specific page
to redirect to.

Closes #1440.


nylira:
Thanks!

PR #1445, opened 2015-11-29
add Apple iWork for iCloud under proprietary Productivity section

### LINKED ISSUE #1444: iCloud appears to have hosted document editing
Looking at icloud.com, it appears that Apple is now offering hosted versions of its iWork suite. If someone can confirm this, we should add it to the proprietary column of the [productivity section][1].

 [1]: https://prism-break.org/en/all/#productivity

vyp:
Closes #1444.


nylira:
Thanks!

PR #1450, opened 2015-12-08
-Redmatrix, +Hubzilla, +Zot

jeroenpraat:
Redmatrix is defunct. The successor is Hubzilla. Also added the protocol Zot that powers Hubzilla.


jeroenpraat:
I'm going to that again, but without the Zot protocol. That will come later, cause I expect some discussion. So lets desperate these two.

PR #1464, opened 2015-12-22
Automatically show product notes in category notes

twiss:
Shows the notes of all the products in a subcategory when viewing the subcategory.

Often, the notes of the products were duplicated in the notes of the subcategory, in which case they were removed there.

I only touched the English locale to make the initial PR a bit uncluttered and also because I didn't know the l10n process - are there any tools in use? Do I just overwrite with the modified strings in other locales, or do I have to "transform" the other strings to match the English modifications?

Advantages:
- Notes no longer talk about products that are not in the subcategory. ([example](https://prism-break.org/en/subcategories/android-instant-messaging/), [example](https://prism-break.org/en/subcategories/servers-productivity/))
- Notes are deduplicated. In a few cases, there were minor differences.

Solvable disadvantages:
- Notes are now ordered by name of the product, alphabetically, just like the actual product cards. (E.g. `/en/subcategories/gnu-linux-web-browsers/`). To solve this, it might be enough to have a binary "recommended" flag which hoists the project to the first spot and also gives it a green or blue recommended label.
- Many subcategories contain "we don't recommend ... because ..." notes, which are now no longer below the product notes. A solution would be to move these notes to hidden "rejected projects" notes, which would also hide them in case they are not applicable on the selected platform.

Disadvantages:
- When a product is in multiple subcategories, we can no longer differentiate the notes per subcategory. (E.g. Thunderbird in `/en/subcategories/gnu-linux-instant-messaging/`, which is probably better than before but still non-ideal.)
- Notes are less flexible and it's harder to get them to say what you want.


vyp:
> I only touched the English locale to make the initial PR a bit uncluttered and also because I didn't know the l10n process - are there any tools in use? Do I just overwrite with the modified strings in other locales, or do I have to "transform" the other strings to match the English modifications?

Basically if they're not already translated (i.e. still english), then we also change them to keep them updated so that any future translators will have the correct up-to-date text in place for them to just translate (no need to compare with the english version). We tend to not worry too much about keeping the ones already translated up to date because there aren't any designated translators or anything (as I understand it), and an out-of-date translation is probably better than no translation.

Sadly there aren't any tools in use that I know of. But at the same time I find that some simple search and replace (or editor macros) do the job sufficiently (if you mean making the same change in multiple files).

I have though used some simple scripts to manipulate the json (in `./source/{db, locales}`) in the past and it's pretty easy. Although I found that the default json module for python was weird in that it didn't retain the order of the keys of the json, so iirc I used node instead. (there's probably a setting to fix that though)

Just a side note, @nylira did mention recently that he was going to try to rewrite the site using rails in an effort to make all this easier somehow (I think), but I'm not sure what the progress has been with that so far.

twiss:
To make this PR not-impossible without throwing away tons of translations, I created a Transifex account for localization: https://www.transifex.com/prism-break/prism-break/. To try it out for one of the languages I updated already, [install the transifex cli](http://docs.transifex.com/client/setup/), `./tx pull`, and `make nl` or `make fi`.

If you think the transifex thing is a good idea, I'll gladly add or transfer admin rights to who you want. If not, I can also just `./tx pull` myself and commit that.

The transifex commit above is very bare bones, it doesn't update the README, etc.

PR #1466, opened 2015-12-23
Fix category links in /categories/

twiss:
https://prism-break.org/en/categories/

I don't think this page is linked to anywhere, I came there by removing the end of the url.


nylira:
Thanks.

PR #1467, opened 2015-12-24
Automatic language selection

### LINKED ISSUE #1448: Detection Of Language Preference
If I visit https://prism-break.org/ it redirects me to https://prism-break.org/en/ even though my languages are set to <code>de-DE,de;q=0.8,en-US;q=0.6,en;q=0.4</code>. The /de/ version looks complete to me so why not redirect there? Same with all the other languages. I don't know how to detect the Accept-Language header but shouldn't be too hard.

twiss:
Fix #1448.

This only looks at the first value in the Accept-Language header, but it should work in most cases.

### DIFF
--- a/source/dotfiles/.htaccess
+++ b/source/dotfiles/.htaccess
@@ -1,83 +1,87 @@
 RewriteEngine On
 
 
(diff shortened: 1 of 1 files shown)


yegortimoshenko:
It seems to be a very good idea. Can this be merged?

Artoria2e5:
Please sort zh-HK/MO into the zh-hant stash with zh-TW too.

Issue #471, opened 2013-08-07
Posteo.de is not free software


ghost:
Remove Posteo.de from e-mail services since it is not free software.

toogley:
What parts are missing here https://github.com/posteo  ?

PR #1473, opened 2015-12-30
added posteo.de

toogley:
added reference to the english project. added reference for translation to non-English projects.

* It seems they engage heavily in privacy things(e.g. their transparency report once a year, in which they cover also data approaches of the german police.)
* Also they have developed an -- in their words -- fully anonymous payment system.


toogley:
It seems posteo.de was someday included but removed, as  [this](https://github.com/nylira/prism-break/pull/477 ) suggests.

hasufell:
https://github.com/nylira/prism-break/issues/471

PR #1480, opened 2016-01-23
Added DNSCrypt.is

russianmonk13:
I added a service I personally host called DNSCrypt.is. It’s a DNSCrypt
enabled, non-logging, uncensored, recursive DNS server with DNSSEC
support based strictly in Iceland. Per the guide I updated all the
*-projects.json and added the logo to
/source/assets/images/logos/medium/.


hasufell:
Where's the license?

Zegnat:
I am not sure we need to start listing DNS providers… we currently list [CloudNS](https://prism-break.org/en/projects/cloudns/) (PRISM Break link) as an external DNS provider with DNSCrypt support. Does DNSCrypt.is offer anything more? Otherwise I would say we should stick with the one we already recommend.

> Where's the license?

DNS providers do not have licenses, as they are simply a service.

hasufell:
> DNS providers do not have licenses, as they are simply a service.

The service runs software and the way this software is licensed can make a huge difference. dnscrypt-proxy (which is apparently used here) is ISC licensed.

fechterjm:
The license is at the following URL https://dnscrypt.is/license.html (in addition a link at the bottom of the side bar). 

The main difference is that DNSCrypt.is is the only DNSCrypt service based strictly in Iceland. The reason this is a big difference is that, from what I've read Iceland has some of the greatest (if not the best) internet privacy laws.

merlinnusr:
If you have Cloudnss(Actual DNS server is in australia only) --why don't put in the list ? to have more providers (In Iceland ).To me is a good aidea .

berrythesoftwarecodeprogrammar:
unrelated to dnscrpyt.is but related to the DNS section:

it says `OpenNIC has not adopted an official policy concerning log query privacy/anonymization. More information here.`. it should probably say the same about dnscrypt as every server has their own features and anti-features. here is a page im working on which aims to let users choose a dnscrypt server more easily https://berr.yt/h/dnscrypt-resolvers/ (i dont want it listed though as its no where near finished). but it just shows what kind of features dnscrypt servers might have or might not. if you wanted to list another resolver it should probably be `soltysiak` and `soltysiak-ipv6` as they have everything `cloudns` has but they also have ipv6 and dnschain support (see here: https://github.com/okTurtles/dnschain/blob/master/docs/How-do-I-use-it.md#Servers )

yegortimoshenko:
This can be closed: seemingly, the service is dead.

https://dnscrypt.is/ provides hoster's certificate:

> dnscrypt.is uses an invalid security certificate. The certificate is only valid for the following names: *.1984.is, 1984.is The certificate expired on January 31, 2018, 7:17 AM. The current time is March 1, 2018, 5:25 PM. Error code: SSL_ERROR_BAD_CERT_DOMAIN

http://dnscrypt.is/ responds with what seems to be default 1984.is index page. http://dnscrypt.is/license.html responds with 404.

soredake:
https://dnscrypt.info/

yegortimoshenko:
> https://dnscrypt.info/

Sure, but this pull request is about DNSCrypt.is the service, not about DNSCrypt the project.

alerque:
Agreed. This PR can be closed, it should not be merged. Even if the service came back online in the near future I'd want to give it some time and another review.

PR #1485, opened 2016-02-07
German translation for Kontalk

webratte:
I don't know if a PR the right way for this. But it was not so much work for me ;-)


webratte:
Thank you ☺

PR #1491, opened 2016-02-28
Add Up1

vyp:
https://up1.ca/ has an onion address, would that count as a 'Tor protocol usage'? I put it there anyway, because I guess it might be, at least allowing it.

Also, not sure if you want to actually remove ProtectedText, hence why it's a separate commit, if you want to cherry pick it, or if you want me to remove it. cc @Zegnat


vyp:
Also assuming slugs can have numbers in them here ("up1")?

Zegnat:
I *think* it can replace ProtectedText completely. But I am still a bit undecided on it. Hopefully some others will chime in. Also two notes:

1. Why is it getting listed as experimental software?
2. Should we rewrite the Riseup Tor services line to include a reference to their hosted Up1 instance at https://share.riseup.net & https://6zc6sejeho3fwrd4.onion?

vyp:
Okay @nylira maybe do not merge straight away to give others a chance to voice any opinions about removing protectedtext.

@Zegnat 1. Because the latest release number is < 1. Do you think I should change it to released?
2. Yes we should imo, good idea. I will do this later.

Zegnat:
> Because the latest release number is < 1. Do you think I should change it to released?

Good catch. Maybe they aren’t using SEMVER. @k3d3 can probably shine some light on that.

vyp:
Yeah I always just try to assume something like < 1 = something like beta or not fully stable, if I don't have anything else to go by that is. That doesn't mean there cannot be exceptions of course. :)

k3d3:
We could probably label it as a 1.0 release, as it is very stable. We were originally using SemVer to indicate that it was beta/unstable, however we've had it running for ~8 months without a single stability issue. We could probably update the version to 1.0 soon to indicate that it's stable.

@andre-d any thoughts you'd like to add, or does that sum it up well?

Zegnat:
> We could probably update the version to 1.0 soon to indicate that it’s stable.

That’s what I thought when I was giving it a look. I also do not think the use of SJCL requires a project to be listed as experimental at this point either. I’ll hold of on merging for a bit until we get that sorted out then.

Let us know what you all over at Up1 decide, and then let @vyp know so he can update the PR :smiley:

andre-d:
The up1 codebase is at a point where we promise backwards compatibility for any client uploading to up1 along with URL backwards compatibility.  We do not yet guarantee backwards compatibility for any clients downloading files from up1.

I am against calling it 1.0 yet as we are experimenting with changes which may require clients who download from up1 to be updated (Such as changes to use a webcrypto algorithm, or file format changes) but from a user perspective hosted clients and servers like up1.ca are stable (although  not audited by a third party aside from a few users we get looking at the github from time to time).

andre-d:
In other words, it is stable from a user perspective, unstable from a downloading/decrypting API perspective

k3d3:
SemVer allows backwards-incompatible changes with major versions though, so this is something we could change with a new major

andre-d:
We can call what we have 1.0 if web crypto experiments or possible changes with galleries are called 2.0, feel free to push a 1.0 tag if you like.

andre-d:
I also would like to remind folks about the experimental and near-hobbyist nature of the up1 project and the importance of using other mature, audited, and staffed projects where security is more than a novelty.

k3d3:
@andre-d @Zegnat @vyp I've tagged the current Up1 codebase as v1.0.

vyp:
@Zegnat I don't know how I missed point 3, but you're right, that should be done too. I will also remove the experimental label based on the discussion. Also, I just realised, I guess the more correct subcategory for Up1 is "Media Publishing", not "Productivity".

vyp:
@Zegnat The list of addresses link in the Riseup services line already has the Up1 instance (at share.riseup.net), so I just added "data hosting (via Up1)" as a service on that line.

k3d3:
You probably had it as Productivity because you were replacing ProtectedText :)

Though it's definitely a bit of both, the main functionality is image hosting with paste hosting as a close second, so it would likely fit better under Media Publishing.

(Can services exist under multiple categories?)

vyp:
@k3d3 

> You probably had it as Productivity because you were replacing ProtectedText :)

Yup. =]

> (Can services exist under multiple categories?)

I see now that it can! (i.e. Can exist under mutiple what is called "subcategories" in the source.) I will amend this later and put Up1 under both the Media Publishing and Productivity subcategories. Probably in a day or two.

vyp:
Apologies for the delay here, I was so busy I forgot to do this.

I added it as a third commit because I currently don't have the time to amend the first commit with this change. Because, unless I'm missing something, that would require redoing the work of the second commit, which isn't difficult, but I just don't have the time to do it right now. So I guess this will have to do for now.

nylira:
Thanks!

vyp:
That was quick! Thanks @nylira.

PR #1504, opened 2016-03-22
Add pass.

mwleeds:
Pass is a simple password manager and pretty much all you need to depend on is a Unix-like command line with GPG installed.


nylira:
Thanks!

themsay:
How about PasswordStore port of Pass made for Android user?

You can check it out from F-Droid platform, very useful password manager that work well Openkeychain in place of GPG.

Hillside502:
For "PasswordStore" read "Password Store"

F-Droid
https://f-droid.org/repository/browse/?fdfilter=PasswordStore&fdid=com.zeapo.pwdstore

Please note beta status.  Same versions on both F-Droid and Google Play.

Password Store (beta) – Android Apps on Google Play
https://play.google.com/store/apps/details?id=com.zeapo.pwdstore

mwleeds:
I use that app and it works well. @zeapo

PR #1510, opened 2016-04-02
Lithuanian translation

techwebpd:
i translated into lithuanian


Zegnat:
I think we would need [DB language files](https://github.com/nylira/prism-break/tree/master/source/db) for Lithuanian too, before this would be useful to merge. Will you also be working on those, @techwebpd?

techwebpd:
i will

Zegnat:
Nice.

Just so you know, I will hold off on merging until all language files are there, to keep build problems at bay.

strugee:
@techwebpd ping? WIll you be able to add the additional data?

yegortimoshenko:
This has been based off Russian translation and is quite stale at this point. That said, it is still worth merging to ease translation for someone who will work on adding Lithuanian locale to PRISM Break in the future. It isn't hooked up to any of the build files, so it should be harmless.

yegortimoshenko:
Superseded by #1945.

PR #1513, opened 2016-04-07
Update french links

xinxinxinxinxin:


Zegnat:
Thanks for localising those external resources we link to!

PR #1518, opened 2016-04-18
Added Umbrella (secfirst)

m-king:


merlinnusr:
wut

strugee:
Umbrella is an interesting project, but unfortunately we need an explicit license in the repository. That legal page is not good enough. Filed securityfirst/Umbrella_android#97. Also, it isn't available in F-Droid:

> We intend to make it also available on the Amazon App Store and through the F-Droid Platform.

Even if those issues were resolved, I'm not sure where and if it belongs on PRISM Break. But it's certainly interesting.

strugee:
@merlinnusr hi! I notice you've been commenting on some PRISM Break issues recently. While we all think it's great that you're here, comments like "wut" don't add anything to the discussion and just generate more email. If you could refrain from comments like that in the future, that'd be great. Thanks, and welcome to the PRISM Break community.

securityfirst:
@strugee Hi, we've added our GPLv3 license to the repository. It would be great to have Umbrella included in Prism-Break. Let us know if there is a best practice way to host it, that works for you. We currently have it on:

F-Droid:
F-Droid Repo: https://secfirst.org/fdroid/repo 
F-Droid fingerprint: 39EB57052F8D684514176819D1645F6A0A7BD943DBC31AB101949006AC0BC228

Google Play: 
https://play.google.com/store/apps/details?id=org.secfirst.umbrella

Amazon:
http://www.amazon.com/Security-First-Umbrella-made-easy/dp/B01AKN9M1Y

rorymbyrne:
Hi @strugee,

If there is anything else you need from us regarding Umbrella - just let us know!

strugee:
Sorry, totally missed this.

I don't know if we have a policy on third-party F-Droid repos, so I don't know if that would outright prevent inclusion on PRISM Break. (Probably yes.) But we'd strongly prefer inclusion on the official F-Droid repo.

That being said, I'm still not sure where this would fit in on PRISM Break, particularly since it's not really an alternative to an existing proprietary service.

rorymbyrne:
Thanks.

Yes I guess weirdly Umbrella doesn't really have an exact direct proprietary alternative (Woo hoo!) for some of the things like security lessons. Though related to the dashboard I think there is a strong argument that a lot of the paid for stuff (non open source) like these is exactly like that:

International SOS Assistance
https://play.google.com/store/apps/details?id=com.infostretch.iSOSAndroid

SOS Assist
https://play.google.com/store/apps/details?id=com.sociable.sos.assist 

Disaster Alert
https://play.google.com/store/apps/details?id=disasterAlert.PDC

Hazard Alert
https://play.google.com/store/apps/details?id=com.hazardalert

rorymbyrne:
On a similar idea. There is quite a lot of proprietary services that do Emergency Provision and Alerts. While there is also a growing amount of open source tools that do it, like Panic Button (https://www.panicbutton.io/). It might be worth looking at opening up a separate category for them?

strugee:
@vyp (@Zegnat...?) opinions?

vyp:
Sorry I have been trying to find time to check out Umbrella, but I couldn't find time. And I still will not be able to for a while a longer. :(

In the meantime, this PR is missing a stub for `ca-projects.json`. Also the "Security" subcategory does not exist, it would need to be added to the `source/locales/*.json` files. Perhaps use a better name than just "Security" though.. but I'm not sure what else to suggest yet.

rorymbyrne:
Hi, just wanted to see if it would be possible to finally get Umbrella into Prism-Break? :)

strugee:
As interesting as Umbrella is I still don't see a place for it on PRISM Break. To me we focus on tools that handle (meta)data in some way (whether at rest or in transit). Umbrella does not seem to fit this. In particular, there's no real proprietary "alternative"; Umbrella exists in more of a vacuum than our other projects.

Given the above I think this should be closed.

yegortimoshenko:
I've read some sections of Umbrella, and unfortunately they give quite a bit of dangerous advice, including that rooting phone makes it more secure. I also think this should be closed, but for a different reason than @strugee.

PR #1531, opened 2016-05-16
Tails support I2P

xinxinxinxinxin:
https://tails.boum.org/doc/anonymous_internet/i2p/index.en.html

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -5870,6 +5870,7 @@
       "SSL/TLS",
       "OTR",
       "Tor",
+      "I2P",
       "XMPP"
     ],
     "categories": [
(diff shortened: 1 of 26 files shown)


alerque:
Please squash these all down to a  single commit for review. Thank you!

xinxinxinxinxin:
Can I do that with the web interface ?

alerque:
Actually I'm not sure if there is a way to do it from the web interface. If it exists I think it would be in the settings for the upstream project so one of the collaborators here would have to enable PR merges with squash options.

Baring that you could close this and open an alternate PR [using this branch](https://github.com/nylira/prism-break/compare/master...alerque:xinxinxinxinxin-patch-5?expand=1) in which I squashed these for you and put the result in a branch on my fork (the commit is still marked as authored by you). Otherwise you'd need to use a git client to manipulate these commits in your fork.

alerque:
See #1532.

PR #1555, opened 2016-07-19
Add 1Password as proprietary password manager.

jornane:
This commit adds 1Password as a proprietary password manager, as suggested and not contested in a comment on #1034.


vyp:
Good catch! One thing, the logo should ideally be 32x32, like the other rejected logos.

jornane:
Pushed new commit. Be adviced that modern displays (e.g. that which Apple calls Retina) will make 32x32 icons blocky in the current design.

vyp:
Oh, why is that? Ok then I will hold off on merging and let @nylira decide if 64x64 should be used from now on.

jornane:
On such displays, logical pixels are displayed by 2x2 physical pixels (apparently the iPhone 6 Plus even uses 3x3), in order to make the picture sharper. For fonts, this works out of the box, but for images you need to keep this in mind when deciding the image resolution; you can't make a picture sharper beyond it's resolution.

There exist frameworks that select the correct image based on the screen type, given that a 2x-version exists, so that devices without such screens don't need to download bigger files.

alerque:
Given that we're dealing with an assortment of project logos and not big graphics, I think defaulting all image resources in the repository from now on to HIDPI sized versions makes the most sense.

jornane:
So which resolution should I include? 64x64?

nylira:
Looks good here. Thanks!

PR #1556, opened 2016-07-19
Add Firefox as a web browser alternative on iOS.

jornane:
Firefox is available on iOS as well, although it's essentially Webkit with a Firefox UI around it. It doesn't share in Safari's syncing, cookies and other state data though.

EDIT: Link: https://www.mozilla.org/en-US/firefox/ios/

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -1800,6 +1800,31 @@
   },
   {
     "development_stage": "released",
+    "description": "Fast, flexible and secure web browser with a vibrant add-on ecosystem.",
+    "license_url": "https://www.mozilla.org/MPL/",
+    "logo": "mozilla-firefox.png",
+    "notes": "Firefox Mobile does not hide your IP address from third-parties by default. If you want to anonymize your IP, consider using Onion Browser.\n\nFirefox sends searches through Google by default. Consider changing it to DuckDuckGo or Startpage for anonymous web searches.",
+    "privacy_url": "https://www.mozilla.org/en-US/privacy/policies/websites/",
+    "source_url": "https://github.com/mozilla-mobile/firefox-ios",
+    "name": "Firefox Mobile",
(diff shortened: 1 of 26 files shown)


vyp:
I didn't know Firefox was available for iOS. I think this is reasonable considering Firefox is already recommended on other platforms and that Onion Browser is the only one currently recommended for iOS.

vyp:
@jornane Could you add this json to all the other json locale files (in that `/source/db` directory)? If you don't know how to do it quickly, then don't worry about it, they will be added later assuming this is merged.

vyp:
Thanks, that LGTM. But I just realised, I think it may be better to add this under mozilla-firefox instead of firefox-mobile. Because the firefox-mobile supposedly specifically refers to firefox for android (or 'fennec'), and hence all the link for firefox-mobile reflect that. Otherwise perhaps an entire new firefox-ios project could be added. @jornane Maybe hold off on making any new changes until @Zegnat or @nylira give their opinion.

jornane:
Since Firefox for iOS is not based on the desktop version of Firefox, I think it's not correct to add it under mozilla-firefox. If firefox-mobile refers explicitly to the Android version, we should add a new project. 

Otherwise, if firefox-mobile simply refers to a version of Firefox that runs on mobile devices (Android, iOS, Windows Phone), we can re-use that project. I'll wait for @Zegnat or @nylira.

nylira:
Hey, sorry for the massive delays, I think its best to add it as a new project.

vyp:
@jornane So let's add it as "firefox-ios" then. (And to prevent confusion, "firefox-mobile" could be renamed to "firefox-android" later.) :)

jornane:
It's been a while, so I don't really remember how this should look. Should a `firefox-ios` be added to all `source/db/??-projects.json`?

vyp:
What you just did looks good to me, ready for merge?

vyp:
@jornane Thank you for noticing this and for your time updating the PR! :sparkles:

PR #1581, opened 2016-10-29
gajim also stores passwords in plaintext

mimi89999:


mimi89999:
@nylira Somebody forgot about that....

PR #1585, opened 2016-11-01
Add missing translation stubs for Android Password Store

vyp:
Ref #1584.

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -99,6 +99,31 @@
   },
   {
     "development_stage": "released",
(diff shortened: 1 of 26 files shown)


nylira:
Thanks

vyp:
Ah right, for the future feel free to open an issue and assign it if you want, to avoid duplicate work. :) Otherwise I'll assume it has already been taken care of like it was here. :+1:

nylira:
I was busy rebuilding the site instead of pushing my latest local version,
my bad. Will let you know if I need translations done.

Issue #1005, opened 2014-06-04
Remove or modify the inclusion of Cyanogenmod due to recent changes in the privacy policy that expand data collection.


gayse:
I understand there has been discussion before on the inclusion of Cyanogenmod (#903, #726, #654), but I believe that this new development warrants a new issue to be created and falls in line with the project guidelines due to a significant change in the development of the software.

Cyanogenmod's privacy policy has been updated recently, and the new privacy policy contains troubling language that I believe goes against the site's mission to promote free and ethical software.

https://cyngn.com/legal/privacy-policy/

https://cyngn.com/blog/2014/06/privacy-policy-and-terms-of-use-update/

> "We may collect information such as occupation, language, zip code, area code, unique device identifier, location, and the time zone where your product or device is used so that we can better understand customer behavior and improve our products, services, and advertising."
> 
> "When you create an account (such as a CyanogenMod Account or a CyanogenMod Forum Account), register your devices or products, purchase a device or product, download a software update, contact us or participate in an online survey, we may collect a variety of information, including your name, mailing address, phone number, email address, contact preferences, and credit card information."
> 
> "Third Parties - As Cyanogen Inc. continues to expand, they enter agreements with the likes of Google and other parties for services seen on retail hardware. Language has been added to reflect both the Terms and Privacy Policy as it relates to third parties."
> 
> "Cyanogen websites, products, applications, and services may contain links to third-party websites, products, and services. Our products and services may also use or offer products or services from third parties. Information collected by third parties, which may include such things as location data or contact details, is governed by their privacy practices. We encourage you to learn about the privacy practices of those third parties."

The privacy policy does not contain any language allowing the user to drop-out of this monitoring, unlike earlier versions of the policy and software. A developmental shift away from privacy and security has been enough to get LMDE removed (#805), and as such I believe the precedent set in that issue should also be applied to this one as well.

Thanks everybody for their time and consideration of this issue, I understand pull requests are preferred here and I will try my hand at creating one shortly.

alerque:
This issue is obsolete and can be closed. Cyanogen has been removed.

Issue #903, opened 2014-03-14
Suggest OMNIrom instead of Cyanogenmod?


wolftune:
Apparently Cyanogenmod is going toward proprietary stuff and various people are suggesting OMNIrom as the next best option to Replicant.

lukateras:
Not up-to-date (unlike LineageOS) and has as many blobs as LineageOS:

https://github.com/omnirom/android_device_samsung_i9300/blob/android-5.1/proprietary-files.txt
https://github.com/LineageOS/android_device_samsung_i9300/blob/cm-14.1/proprietary-files.txt

If we are going to tolerate blobs running on the main CPU, LineageOS would undoubtedly be a much better pick.

strugee:
Yeah, this issue predates LineageOS by many years. I don't see the point in keeping it around now.

Issue #723, opened 2013-11-06
CyanogenMod contains Google Tracking. Please down it.


gitkitten:
Many Request fired from Cynamon 9 , without user notice:
Sample (I hide some numbers for my privacy, sorry about that)

www.google-analytics.com/__utm.gif?utmwv=4.6ma&utmn={XXXXX}&utmcs=UTF-8&utmsr=540x960&utmul=en-US&utmp=%2FROM_Manager&utmac=UA-{YYYYYY}-3&utmcc=__utma%3D{ZZ}.{ZZ}{ZZ}{ZZ}.{ZZ}.{ZZ}.{ZZ}.1

{XXXXX} ... numbers
{YYY} ... numbers
{ZZ} ... numbers

So I want to down its rank.

"with either Cyanogenmod or Replicant, safeguard your data..."
->
"with either Replicant or Cyanogenmod, safeguard your data..."

and

CyanogenMod
Replicant
Firefox OS
->
Replicant
Firefox OS
CyanogenMod
# I don't use ANY GOOGLE, because I HATE IT
# Of course I didn't install Gapps
# I don't install ANY APKs, just email app(K) and SIP app(from guardianproject.info)

And another problem...
# CSipSimple or Chatsecure connect to httpS://guardianproject.info/

without user notice.
What the FUC\* is this? So I block him with my DNS for now. Apps runs with no problem.
# CM9 doesn't honor hosts file. Still try to connect traitors(such as Google)...

adb pull /system/etc/hosts hosts
adb remount rw
adb push hosts /system/etc/
adb shell chmod 644 /system/etc/hosts
adb remount ro

^ I use this command to:
Android 2.x + ROOT ... Obey hosts file
Cynamod 9 (root) ... Doesn't work. Seriously, if someone can do this, would you tell me how?

And also, how about adding a guide such as:
"After you replace a ROM, update hosts file to prevent your phone connect to Google."

I can share my collection of tracking+ads hosts file, if someone need it. :)

Issue #654, opened 2013-09-23
CyanogenMod has recently gone commercial


istubbs:
Should CyanogenMod still be listed now that they have gone commercial?
https://plus.google.com/106978520009932034644/posts/L8FJkrcahPs
With them now dual-licensing code to allow for closed-source alterations that will not be committed back to the public repository as I read it, CM suddenly seems a lot less secure to me.

alerque:
This issue is obsolete and can be closed, Cyanogen was removed.

Francewhoa:
For those not familiar with what happen with CyanogenMod. CyanogenMod went commercial. In other words, Cyanogen Inc. tried to monetized the open source community project CyanogenMod. Sounds like Cyanogen Inc. were serving money first instead of serving the community first ;) As a result the community forked CyanogenMod to LineageOS at https://github.com/lineageos and http://lineageos.org Which is fully community-owned. And without strings attached to any corporations :) So really free of corporate interference.

On 23 December 2016, Cyanogen Inc. announced that they were shutting down the infrastructure behind CyanogenMod. This was shortly followed by news that the main CyanogenMod project would migrate, renaming itself as "LineageOS". On December 24 2016, Head of Developer Relations and community forum administrator Abhisek Devkota, a Cyanogen "core team" member, wrote that the community had lost its "last remaining advocate" within the company and its voice in Cyanogen Inc. and its software's future. He stated that while "that this most recent action from [Cyanogen Inc.] is definitely a death blow for CyanogenMod", the community had already begun taking the steps needed to fork the project under a new name, and aimed to return to its grassroots origins while retaining professional approaches adopted during the Cyanogen Inc. era. Due to the negative connotations attached to Cyanogen Inc's conduct, as well as the scope for legal dispute, the forked project decided not to use the existing brand names "Cyanogen" or "CyanogenMod", which in any case belonged to the company.

Homepage http://lineageos.org/ 

Read more https://en.wikipedia.org/wiki/LineageOS 

Related post https://web.archive.org/web/20161225144318/https://www.cyanogenmod.org/blog/a-fork-in-the-road

alerque:
@Francewhoa Please don't double post, it just makes more work for those of us monitoring this project and dilutes relevant discussion across multiple locations. As this issue is already resolved, anything relevant to an action point (such as possibly adding Lineage) should stay on the Lineage issue.

Francewhoa:
alerque :) I agree. I tried to remove my duplicate post. But can't find it.

PR #1588, opened 2016-11-03
Move CyanogenMod to proprietary section

hasufell:
https://github.com/nylira/prism-break/issues/654
https://github.com/nylira/prism-break/issues/723
https://github.com/nylira/prism-break/issues/726
https://github.com/nylira/prism-break/issues/903
https://github.com/nylira/prism-break/issues/1005


nylira:
Thanks!

Alexay:
@nylira Are you guys sure you're not mixing up the community-developed free and open source CyanogenMod with CyanogenOS? The links provided in those issues all pertain to CyanogenOS.

hasufell:
see http://redmine.replicant.us/boards/33/topics/1041

Alexay:
@hasufell that was 3 years ago. They have since separated the proprietary version into CyanogenOS and kept CyanogenMod free and open.

Hillside502:
Commercialization controversy - CyanogenMod - Wikipedia, the free encyclopedia
https://en.wikipedia.org/wiki/CyanogenMod#Commercialization_controversy

hasufell:
> They have since separated the proprietary version

I don't see any information that CyanogenMod stopped using/shipping proprietary drivers. And wikipedia also clearly states:

> Replicant is a CyanogenMod fork that removes all proprietary software and drivers and thus avoids all aforementioned legal issues. However, Replicant does not support devices which depend on proprietary drivers, which is most phones as of 2016.

mimi89999:
@hasufell 
> I don't see any information that CyanogenMod stopped using/shipping proprietary drivers (which would render it as useless as replicant).

Yes, it does. As half of the GNU/Linux distribution, that prism-break is recommending. Where is the logic? I don't understand that...

hasufell:
> As half of the GNU/Linux distribution, that prism-break is recommending. Where is the logic? I don't understand that...

huh, ever heard of kernel deblobbing? Some distros we recommend do that per default (Parabola I believe), others allow it rather easily (Gentoo, ArchLinux). And: deblobbing is a script you can use on any linux distro prior to building your kernel. Can you point me to a similar technique on Android?

mimi89999:
@hasufell 
> ever heard of kernel deblobbing?

Yes, Linux-libre has a script for that...

> Some distros we recommend do that per default 

Yes, some do (Debian, Fedora, gNewSense, Parabola) and other don't.

> And: deblobbing is a script you can use on any linux distro prior to building your kernel. Can you point me to a similar technique on Android?

Theoretically you could run the Linux-libre script searching for blobs, remove them and rebuild the kernel with Cyanogenmod, but nobody would do that since without them and proprietary drivers and firmware, the device would be a brick.

Also, let's take as an example OpenSuse, that you recommend. It has both, free and proprietary software, firmware, drivers in it's repos...

hasufell:
> Theoretically you could run the Linux-libre script searching for blobs, remove them and rebuild the kernel with Cyanogenmod, but nobody would do that since without them and proprietary drivers and firmware, the device would be a brick.

What? That's not even the same kernel.

Issue #1592, opened 2016-11-07
Correct Ixquick links


IDKwhattoputhere:
`ixquick.com` is a mirror of `startpage.com`. The actual Ixquick metasearch engine is located at `ixquick.eu`.
https://www.ixquick.com/eng/ixquick-merged-with-startpage.html
This is obvious if you compare the two Ixquicks.
![](https://cloud.githubusercontent.com/assets/16030872/20043925/22e20dfe-a494-11e6-8309-3bec750dc52f.png)
![](https://cloud.githubusercontent.com/assets/16030872/20043926/22e410d6-a494-11e6-8105-c270ed6c7e75.png)
I don't think linking to the same search engine twice is intended so I created https://github.com/nylira/prism-break/pull/1593. If I'm however wrong feel free to close this.
One more note: `ixquick.eu` seems to direct to a US IP while `eu.ixquick.eu` directs to a EU IP.
![](https://cloud.githubusercontent.com/assets/16030872/20044029/2571dd64-a495-11e6-92b8-db98a7643c1b.png)
Not sure if the URLs should be edited to address this.

alerque:
I agree we only need one of these now.

strugee:
Looks like this should be closed since #1593 got merged?

IDKwhattoputhere:
Not quite yet. I noticed that there are still references to ixquick.**com** in `/source/locales/` and `/source/db/`. Feel free to make a pull request or I will as soon as I have the time.

vyp:
Closed by #1625, thanks @IDKwhattoputhere!

PR #1625, opened 2016-12-03
Correct remaining Ixquick links

IDKwhattoputhere:
https://github.com/nylira/prism-break/issues/1592#issuecomment-262667314


strugee:
@IDKwhattoputhere can you squash these commits into one and fix the message? Otherwise LGTM

PR #1627, opened 2016-12-06
Make localizations inherit from English JSON data

strugee:
WIP for feedback; do not merge. /cc @nylira because most of this was written by you.

This change makes it so that when building localized pages, the English project data is used as a default if a localized version is not provided. This means that when adding new projects, contributors can update the English data and be done with it, as opposed to patching all 23 language files. Updates to existing projects also only need to patch files that have already been localized.

One big downside to this is that it's no longer quite as obvious what's not localized. I have a plan to add some Make targets which will provide a list, per-language, of what still needs to be added. This will make localization via the GitHub web interface slightly more challenging, but I think it's worth it.

However! This implementation is currently broken. You can see the problem by cloning the source branch and running a build:

```
% make es
rm -rf tmp/*
mkdir -p tmp/assets/css tmp/assets/js
cp -r source/assets/fonts tmp/assets/fonts
cp -r source/assets/icons tmp/assets/ico
cp -r source/assets/images tmp/assets/img
./node_modules/.bin/lsc ./source/functions/build/site-es.ls
Failed at: ./source/functions/build/site-es.ls
TypeError: string.replace is not a function
    at slugify (/Users/alex/Development/github/prism-break/source/functions/slugify.ls:5:19)
    at /Users/alex/Development/github/prism-break/source/functions/sort.ls:27:13
    at /Users/alex/Development/github/prism-break/node_modules/prelude-ls/lib/List.js:15:19
    at /Users/alex/Development/github/prism-break/node_modules/prelude-ls/lib/List.js:665:42
    at slugifyList (/Users/alex/Development/github/prism-break/source/functions/sort.ls:24:10)
    at slugifyProject (/Users/alex/Development/github/prism-break/source/functions/sort.ls:40:32)
    at create (/Users/alex/Development/github/prism-break/source/functions/write.ls:316:12)
    at writeProjectsShow (/Users/alex/Development/github/prism-break/source/functions/write.ls:344:19)
    at writeLocalizedSite (/Users/alex/Development/github/prism-break/source/functions/write.ls:21:3)
    at buildSite (/Users/alex/Development/github/prism-break/source/functions/build.ls:42:10)
make: *** [build_es] Error 1
```

Note that for Spanish in particular I've committed a change (to be removed before merge, obviously) that reduces the data to a minimal testcase - Firefox for iOS is the problem and FreeBSD is left in there as a reference for something that works, for whatever reason. :/

Basically, the problem is that at some point in line 15 of `source/functions/sort.ls`, `slugify` gets passed `{ name: 'Web Browsers', slug: 'web-browsers' }` instead of (presumably) `Web Browsers`. Seems like `slugify` is getting called on some data twice somewhere, though for the life of me I can't wrap my head around how.

If people want to look at the code and spot the bug, that'd be great. I certainly can't, at least so far :P

Feel free to commit directly to my branch, although I might squash your commit into one of mine in the course of cleaning this up for merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ LANGUAGES := $(notdir $(basename $(wildcard source/locales/*.json)))
 ASSETS := $(notdir $(wildcard source/assets/*))
 
 # Mark all rules that don’t actually check whether they need building
-.PHONY: default test init reset all $(LANGUAGES) assets css html html_% clean watch watch_css sync
+.PHONY: default test init reset all $(LANGUAGES) assets css html html_% clean watch watch_css sync localize_%
 
 # Turn on expansion so we can reference target patterns in our dependencies list
 .SECONDEXPANSION:
@@ -105,6 +105,10 @@ clean:
 public/.htaccess: source/dotfiles/.htaccess
 	mkdir -p $(dir $@)
 	cp $< $@
+
+# Localizations
+localize_%:
+	lsc ./source/functions/find-missing-localizations.ls $*
 
 #----------------------------------------------------------------------
 # CONVENIENCE FUNCTIONS
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Welcome to the PRISM Break project. Here's a quick overview of the code. JSON co
 
 The prism-break build process relies on several npm packages. Make sure to have [node.js](https://nodejs.org/) installed on your system if you want to contribute to the code.
 
-If you'd like to translate the project to your favorite language, there's no need to install node.js or even download the code. Just edit the appropriate JSON files on GitHub and submit a pull request.
+If you'd like to translate the project to your favorite language, there's no need to install node.js or even download the code. Just add translated data to the appropriate JSON files on GitHub and submit a Pull Request - if anything's missing, the English version will be used instead. That being said, setting up the project locally will allow you to run `make localize_LANG` (where `LANG` is a language code) to get a nice list of things for you to translate.
 
 More information for translators can be found in [`CONTRIBUTING.md`](CONTRIBUTING.md).
 
--- a/package.json
+++ b/package.json
@@ -4,10 +4,12 @@
   "description": "A directory of free privacy-oriented software.",
   "main": "Makefile",
   "dependencies": {
+    "es7-array.prototype.includes": "^2.0.0",
     "graceful-fs": "~4.1.4",
     "jade": "~1.11.0",
     "livescript": "~1.5.0",
     "lodash.assign": "^4.2.0",
+    "lodash.isequal": "^4.4.0",
     "marked": "~0.3.5",
     "mkdirp": "~0.5.x",
     "moment": "^2.17.1",
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "graceful-fs": "~4.1.4",
     "jade": "~1.11.0",
     "livescript": "~1.5.0",
+    "lodash.assign": "^4.2.0",
     "marked": "~0.3.5",
     "mkdirp": "~0.5.x",
     "moment": "^2.17.1",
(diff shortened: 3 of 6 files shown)


strugee:
@nylira ping? I was really stumped when I sent this PR; I was hoping you could take a look and maybe figure out what's going on

nylira:
Hey thanks for the work! I'll take a look at the issue you're having later today.

strugee:
:thumbsup: thanks!

nylira:
I've narrowed down the bug to the string `"Firefox Mobile",`. It seems like changing Firefox to anything else "Furefox", "Firefoxe" fixes the problem. Changing Mobile to anything else "Mobil", etc. fixes the problem. It's just the exact string "Firefox Mobile".

I'm not sure what causes this. However, the solution I'm going with is renaming Firefox Mobile to Firefox for iOS and Firefox for Android. It's a clearer name anyway. I'll push these changes to the main repo and you can sync your changes from there.

vyp:
@nylira Yep sorry I was supposed to rename firefox-mobile to firefox-android before anyway but I forgot!

Makefile
```diff
@@ -159,3 +160,56 @@ sv: clean_tmp mkdirs copy_assets build_sv finalize
 tr: clean_tmp mkdirs copy_assets build_tr finalize
 zh-CN: clean_tmp mkdirs copy_assets build_zh-CN finalize
 zh-TW: clean_tmp mkdirs copy_assets build_zh-TW finalize
+
+# Localizations
+
+localize_ar:
+	$(LIVESCRIPT_BIN) $(LOCALIZE_SCRIPT) ar
```
alerque: This entire mass of localization targets could be simplified down to two lines:
```Makefile
localize_%:
        $(LIVESCRIPT_BIN) $(LOCALIZE_SCRIPT) $*
```

strugee: Thanks! Didn't know Make could do that. Will send a PR simplifying the existing build targets too

alerque: Ouch, I'd never seen the whole Makefile, only this diff. Why don't you let me take a poke at it, there are actually several things that could be fixed.

strugee:
@nylira before I forget - do you prefer `"`s or `'`s? Or does it matter? The existing code is a mix of the two

strugee:
Alrighty, the core of this PR works pretty well. However I'd like to add some checks to help us identify existing duplication that can be removed. I've started work on this, but it's buggy currently.

strugee:
OK, I think this is ready for merge! If anyone wants to review the code here that'd be awesome, otherwise I'll just merge it in a couple days

Makefile
```diff
@@ -121,3 +121,8 @@ watch_css:
 sync:
 	rsync -azru --delete --stats public/ ../prism-break-static/public/
 	(cd ../prism-break-static; git add -A; git commit -m 'regenerate'; git push)
+
+# Localizations
+
+localize_%:
+	lsc ./source/functions/find-missing-localizations.ls $*
```
alerque: This either needs to me added to the PHONY list or reworked so that the target is a file path. It looks like it is intended to be PHONY.

alerque: Also I would move this up to into the 'FUNCTIONS' section.

Sorry I haven't looked through this too much, but what output is  expected from this target? Should it be run every time language X is run? If so it should be added as a dependency to the `html_%` target, but we need to think about the order too. Does it have to run first or can it run any time?

strugee: @alerque also, to answer your question directly, it doesn't write anything to disk - just outputs to the console. Therefore it can be run at any time.

alerque: Alright in that case maybe we call it from Travis based on what languages are touched in a PR, but that's a job for another issue. For this one it looks like all you need to do is add this target to the `.PHONY:` list (since isn't something that the build process is expected to output and hence there is no possibility of dependency checking or knowing whether it's 'needed' at make time).

strugee: @alerque I did, but forgot to push :P

Done for real now!

strugee:
Merging since there hasn't been activity for a couple days

PR #1636, opened 2016-12-14
Simplify Makefile build targets

strugee:
Thanks to @alerque for (indirectly) pointing this out in #1627 :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,58 +43,8 @@ watch_css:
 build_css:
 	$(STYLUS_BIN) $(STYLUS_PARAMS) <$(STYL) >$(CSS)
 
-build_ar:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)ar.ls
-build_ca:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)ca.ls
-build_de:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)de.ls
-build_el:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)el.ls
-build_en:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)en.ls
-build_eo:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)eo.ls
-build_es:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)es.ls
-build_fa:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)fa.ls
-build_fi:
-	$(LIVESCRIPT_BIN) $(BUILD_DIR)fi.ls
-build_fr:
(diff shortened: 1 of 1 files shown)


alerque:
This is good as far as it goes. There are a number of further things that could be cleaned up through. I'd never looked at the Makefile before but it's apparent somebody was just blundering around with `make` (no offense meant, it does lend itself to that). I don't know you want to merge this or wait for me to add some more commits to it, but I'll work up a few things.

One question through ... is it reasonable to expect `jq` or  some other command line JSON tool to be a build time requirement?

strugee:
@alerque feel free to send PRs to strugee/prism-break against the `simplify-build-targets` branch - I'll merge them straight away and they'll show up here.

I've always felt that it would be better to just get rid of the Make dependency and use a JS-based build tool. Then we can tell people to `npm install` and e.g. `npm run build-en`, and it would Just Work™, even if we add new deps.

@nylira what do you think? I mean, this should probably be merged either way, but before we discuss/@alerque works on solving Make suckage with a new build dependency, I think we should consider just dropping Make entirely.

(Apologies since that's a tad incoherent - it's 1:15 AM here :P)

nylira:
Going with a pure JS build process is fine. Go for it!

nylira:
If jq or rq can be installed via npm that would work. I want to keep the build process portable and multiplatform.

alerque:
Personally I despise JS based build systems. If that's the way it's going to go, you get somebody else to program it. As archaic as `make` can be sometimes it's close ties to the shell and general Unix way of doing things and number of build system optimizations and general know-how it brings to the table isn't really matched by nearly anything else. Sure you _can_ do it all in Grunt or whatever, but you end up having to code a lot of stuff that could otherwise just be left off.

If the `Makefile` is going to stick around I'll clean it up a bit more to be DRY and so that the targets don't need so much manual intervention to, say, add a language. Also I'll fix it so that it actually knows what to build instead of just force-building everything based on how it was called. This should speed up the process a lot by not wasting time building things that are up to date.

hasufell:
> isn't really matched by nearly anything else

And the ways to mess up aren't matched by anything else either (speaking from 4+ years of distro packaging experience).

I think for a project as small as this, it really doesn't matter.

alerque:
Touché.

And by the way the answer to the earlier question is yes, `jq` can  be used as an NPM dependency so it would stay self contained. There are a number of things that could  be done with that tool to make managing the JSON sources easier than doing it all by hand.

strugee:
To be clear @alerque, I was saying that I thought we should go with a JS solution if we needed a bunch of new dependencies to make Make (ha!) work, and I didn't want to object _after_ you'd done a bunch of work. I wasn't suggesting that you specifically should make that happen.

In any case, since we're not adding any non-npm reps, I have no issues sticking with Make :)

PR #1638, opened 2016-12-16
Remove bogus content outside of JSON in ar language data

alerque:
I'm working on a complete overhaul of the Makefile and ran into not being able to build all the languages. It looks like this came from 00be7c1 — somebody must not have actually tried building this 😉 .

(The good news is with the new Makefile and proper dependencies a failed make knows what language didn't finish and `make all` picks up where it left off!)

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -6850,4 +6850,4 @@
     ],
     "slug": "android-password-store"
   }
-]r
+]


vyp:
Thanks! Very nice work with the Makefile being able to detect the point of failure. I don't see a `00be7c1` anywhere though..? But I can see it was added in https://github.com/nylira/prism-break/commit/f93827688e1a08f07d5aaa6f85456b69de2d8e91#diff-0ed0a13851018c8b5c131c81226021a1R6878.

alerque:
Oops, I must have copied the wrong hash, that's the commit I found too.

PR #1639, opened 2016-12-16
Make make smarter

alerque:
Reviewing PR #1627 brought it to my attention that the Makefile for this project was a mess. PR #1636 picked off one bit of low hanging fruit which helped it some with not being so repetitive, but it barely touched the surface. There was still a lot of repetition. That would have been easy enough to factor out with a proper variable list, but looking into that made it all the more obvious `make` wasn't actually being used for any of its features. Basically the entire thing was just a wrapper for a couple short shell sequences. _ALL_ of the targets were phony, meaning there was no possibility for any dependency checking.

The commit messages for this tell the story of refactoring it. Basically this is a whole new `Makefile` from scratch, but I documented each major step. If there are questions about why I did something we can deal with them piece meal. If I need to change bits of this to suite others' usage I can do that and rebase it, so please chime in with reviews.

The basic idea here is to actually use GNU Make's dependency checking features. To do this it needs targets that are actually the generated output files and the inputs used to create these files are listed as dependencies. With this basic pattern, make will know after a change what needs to be rebuilt. As a proof of concept I added a full project watch target (you'll need `entr` on your system) that will rebuild just what needs doing when you update any files in the source (sort of like `stylus`'s watch does for just the css part). Run `make watch`and then change anything in the source directory to check it out.

Otherwise most of what's there should be either self explanatory or –as in cases where I used features many might not be familiar with– documented in comments.

Along the way I made two structural changes to the static site process that seemed to be there for legacy reasons and were just making the build process unnecessarily hard to maintain.

1. I used the same names for asset folders in the output as in the source. Why "icons" in the source was being transformed to "ico" etc. was beyond me. If there is some reason for this, please let me know what it was so we can handle it correctly.

2. I dropped the concept of a "tmp" folder during the build process. The publishing process seems to involve copying to an external folder anyway and nothing was happening in the tmp space that wasn't being copied *carte blanche* to the public directory anyway, so it seemed like an exercise in futility. Again if there is some use case for this, knowing what it was would help me propose a proper solution.

Finally there are a few further things that could be done, but I'm out of time for today and don't want to work on them before getting some feedback on this anyway. One logical next step for example is to ditch the absurd repetition in `source/functions/build/*`. The locale value could just be passed in by recipe instead of needing an otherwise identical script per locale.


Makefile
```diff
@@ -1,111 +1,115 @@
...
 
-# Binaries
-BIN = ./node_modules/.bin/
-LIVESCRIPT_BIN = $(BIN)lsc
-LIVESCRIPT_PARAMS = -cob
-STYLUS_BIN = $(BIN)stylus
-STYLUS_PARAMS = -c -u ./node_modules/nib/
-STYLUS_WATCH_PARAMS = -c -w source/stylesheets/screen.styl -u nib -o public/assets/css/
+# Collect list of static assets that get copied over
+ASSETS = $(shell ls source/assets)
```
hasufell: `ls` is primarily for human readable output and should be avoided in a scripting/programming context. I'd either use `find` or shell globbing/expansion. Here's also one example: https://www.gnu.org/software/make/manual/html_node/Shell-Function.html `files := $(shell echo *.c)`.

Also, debatable, but for variables set by shell functions, I'd usually assign them via `:=` (as the example above does) to avoid potential race conditions (although rather unlikely and not the case here afais).

alerque: True story. I'll fix those points. I actually had a find here than failed for other reasons and I hacked than it to see what was wrong with the find, then forgot about it.

alerque: To my knowledge only directories are expected there. This would adapt to new or renamed directories with no change to the `Makefile`, but if people start putting random stuff of other types it might need some adjustment. See 74eb936 for how I skinned that cat slightly differently before seeing your suggestion, but I think I prefer not even shelling out there and just using internal functions.

alerque:
I just caught a rebase mistake where accidentally removed some targets from the phony list when I meant to re-order them. I just rebased to fix that and take advantage of the other PR that fixed the data.

alerque:
With the latest commit that marks the dependencies as being order specific where appropriate (and the bug fix I made adding `clean` to the phony list) you can also now build languages in parallel: `make -j 6 reset` will build everything from scratch six languages at a time. This trims quite a bit of  time off a full compile if your system has a lot of cores and ram to work with.

nylira:
Thanks for your work. It runs well here, and it's way better than before.

nylira:
If you want to make more changes to the build process feel free to open another request.

PR #1640, opened 2016-12-17
Update dependencies

alerque:
I tested each of these  updates separately building on my machine and none of them seemed to disrupt the process or produce different output. It's a little hard to be sure since we don't have a complete output test suite, but all seems copasetic.

LiveScript dropped the capitals and normalized on lowercase in the NPM registry.


nylira:
Looks fine here. Thanks.

PR #1641, opened 2016-12-17
Unicode entities

alerque:
✔ We encode the JSON files as UTF-8.
✔ We add UTF-8 encoding headers to the HTML.
✔ We write most of our translations using Unicode alphabets.
✔ We serve the site with UTF-8 content headers.
✗ Yet we have a whole gaggle of random characters encoded using HTML escapes from a decade ago that are only needed when serving files encoded with more limited character sets.

I mean what happened to Polish with:

    otwarto&#378;r&oacute;d&#322;owa sie&#263; spo&#322;eczno&#347;ciowa

being used where:

    otwartoźródłowa sieć społecznościowa

would do? Surely that's easier to translate and edit for anybody with a Polish keyboard. The same goes for a half dozen other languages that had exceptionally messy strings.

This PR picks off all the random character entity escape sequences I could find and replaces them with the Unicode code points. The only ones left are:

* `&amp;` which is required because `&` itself denotes the start of an escape sequence in the HTML spec (still). Most browsers are tolerant of this, but there is no reason we shouldn't follow the spec and encode it properly.
* `&lt;` and `&gt;` which encode `<` and `>` and would otherwise be interpreted as HTML tags.

Todo:

- [ ] Get somebody to check my fixes to the Hebrew ordering. As far as I can tell the previous layout order marks were just bogus and not working as intended. My change from a left-direction-mark to a right-direction-mark separates the punctuation from the LTR words so that the punctuation renders with the appropriate sentence instead of attached to a word in the next sentence.
- [x] Make sure right-single-quote makes sense typographically for Italian where it was substituted in place of `&#39;`. I'm pretty sure this is common usage, but if a typographic right single quote is better we could do that too.


strugee:
@alerque:

> Make sure right-single-quote makes sense typographically for Italian where it was substituted in place of &#39;. I'm pretty sure this is common usage, but if a typographic right single quote is better we could do that too.

You seem to have made a typo, since you mentioned "right single quote" twice but didn't mention a different alternative.

alerque:
@strugee The diffenence is between a **typographic** right single quote (`’`) and a regular straight one (`'`). My wording was confusing and I should have called the former a curly quote, but those are the options.

strugee:
@alerque OK, I _kinda_ understand what you're saying. FWIW my dad speaks Italian and I asked him to take a look at the relevant commit:

> [...] at first glance seemed okay, but hard to read on my phone. Italians routinely mix English jargon in Italian sentences. I'll take another look.

strugee:
Follow-up comment from my dad:

> I looked at the GitHub link. The character replaced by &#39 is an apostrophe and is being used correctly. I can't read the entire line because it's masked by the window border and the scroll control is off screen. There do appear to be some typos on lines 259 and  2389 where the e' (accent on top of letter - I don't know how to type them with my keyboard) should be a'. I can't comment on the other language, but it seems reasonable as far as my knowledge goes.

Zegnat:
Most of the quote conversions I am seeing in the Italian file are for contractions, e.g. <i>alla email</i> → <i>all’email</i>. Those should definitely be written with a typographically correct (“curled”) apostrophe, and not be substituted with straight typewriter quotes.

The one non-contraction use is when it uses [short year notation](https://github.com/nylira/prism-break/pull/1641/commits/18b39b1a0f33b32cd4451317b554cc9bde8deaa0#diff-7770fbc536dd7daaa874e616b6d81f5aL1225), this too should not be done with the straight typewriter quote.

In other words, all of the `&#39;` you replaced with `'` in 18b39b1 can instead be replaced with `’`.

alerque:
That's actually what I did originally @Zegnat, then I looked around at some Italian texts and it seemed like the majority usage was to use straight typewriter quotes. Hence why I  added that as something than needed follow up.

@strugee Thanks for checking with a speaker! It sounds like the issues found are not so much to do with the apostrophe type and encoding as they are general issues with the Italian translation. We can certainly fix those but probably doing it in a separate branch from my encoding shenanigans would make sense. It's just easier to test this and make sure I didn't change anything but encoding if I don't also roll in other changes. It also isn't quite clear to me whether he knew he was looking for the difference between quote mark types.

Zegnat:
> <i>[…]</i> I looked around at some Italian texts and it seemed like the majority usage was to use straight typewriter quotes.

Did you look at online texts? Because those are terribly biased towards typewriter quotes. In any language. Mostly because of Windows not giving easy access to the correct symbols. It also does not help that Wikipeda prohibits “typographically correct quotation marks” [[1]](https://en.wikipedia.org/wiki/Wikipedia_talk:Manual_of_Style_(Quotation_marks_and_apostrophes)) [[2]](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style#Apostrophes).

Looking at [Italian books published after 1900 that have full scans available on Google](https://www.google.com/search?lr=lang_it&tbo=p&tbm=bks&q=l%27arte&tbs=,bkv:f,cdr:1,cd_min:Jan+1_2+1900,cd_max:Dec+31_2+2100&num=10), most seem to use correct apostrophes. If we decide to be typographically correct, I vote for using them in Italian and every other language.

strugee:
This has been sitting here for a while. I'm thinking we should probably just go ahead and merge it.

alerque:
I'm going to have to rebase it since it didn't get merged before more strings got messed with.

strugee:
:thumbsup: ping me when you're done

alerque:
@strugee Can you assign me to this so I don't have to keep this tab open to not forget to do this when I have a chance?

alerque:
@strugee I've rebased this PR on the current master (that was hairy!) and also touched up the quotes in Italian per discussion  with @Zegnat. This should be ready to go again. And the sooner it gets reviewed and merged the less likelihood other people will base new changes off the  old encodings.

strugee:
LGTM, thanks!

PR #1644, opened 2016-12-24
Update zh-CN.json

simonasd:


source/locales/zh-CN.json
```diff
@@ -14,8 +14,8 @@
   "Email Alternatives": "电子邮件的替代品",
   "Email Clients": "电子邮件客户端",
   "Email Encryption": "电子邮件加密",
-  "Enterprise Suite": "Enterprise Suite",
-  "enterprise-suite-notes": "The enterprise suite category is for solutions for organizations that cover more than 10 categories in an integrated fashion (ex.: logins work throughout all apps, etc.)",
+  "Enterprise Suite": "企业套件",
+  "enterprise-suite-notes": "“企业套件”分类中列出了一些适合公司或组织使用的软件套装，往往是包含了10种以上不同种类工具的合集（大都支持统一登录等特性）。",
```
alerque: Is the leading curly double quote here intentional? It doesn't seem to end anywhere. I'm not very up on Chinese typography but this doesn't seem quite right.

strugee: @simonasd ping? If you answer the above question I'd be happy to merge this

paperbag: @strugee 
**”** is end quote in Chinese.

paperbag: It needn't double quote. However, it is no mistake in chinese.

alerque: It appears I just didn't see the `”` mark at all, perhaps because my eyes are used to finding it with whitespace breaks around it.

strugee:
LGTM, merging

source/locales/zh-CN.json
simonasd: guys, sorry for my disappear these months. i have some bug to fix in my real-life :( .

`“”` is quote in Chinese, used in pair. You can use it to quote the word you are defining, just like the bold word at the beginning of every Wikipedia entry.

PR #1665, opened 2017-01-10
Add mIRC to Proprietary

kompowiec:


source/locales/pl.json
```diff
@@ -228,12 +228,12 @@
   "Chinese (Simplified)": "Chiński uproszczony",
   "Updated": "Zaktualizowane",
   "By": "przez",
-  "Donate Bitcoin": "Bitcoin wspomóż akcję",
+  "Donate Bitcoin": "Dobrowolna dotacja BitCoin",
```
strugee: What was the problem here?

strugee:
@kompowiec thanks for contributing to PRISM Break! It'd be great if you fixed your commits so there was a single commit adding mIRC and a single commit updating translations, with a commit message like "Updated such-and-such translation."

You can do this using `git rebase -i`. If you need help I'm happy to provide assistance, or if you really can't figure it out I can just push to your branch, if that's okay?

hasufell:
> or if you really can't figure it out I can just push to your branch, if that's okay?

You can rebase locally and push the rebased thing to this repo.

source/locales/pl.json
kompowiec: previous construct was unnatural. Currently reads as follows: voluntary dotation BitCoin.

kompowiec:
I remember that over 4 years ago: https://github.com/kompowiec2/prism-break project code project was very different (use just HTML+CSS) now is TypeScript so i could do a little blindly.

I try insert code in appropriate places and again code review.

source/locales/pl.json
kompowiec: however, if be in accordance with original, can add: wspomóż akcję POPRZEZ BitCoin. Word order is now corrctly with Polish lang.

alerque:
@kompowiec I dont' think the request here had anything to do with where code ended up or the languages involved. Only the _git commit sequence_ was in question with two commits needing to be merged into one. If that doesn't make any sense to you it's actually something easy enough for somebody else do do. I don't have write access to this repo but I think @strugee can squash the commits and push it back to this branch easily enough.

source/locales/pl.json
strugee: Cool. I trust that you've made an improvement, I was just curious.

strugee:
@kompowiec yes, what @alerque said is correct. If you don't mind me pushing to your branch, I can correct the commits for you.

strugee:
Also, the source is written in LiveScript, not TypeScript ;)

kompowiec:
aa i now understand - I do commits from web because can't clone repo local, proably with reason little files and dangerously low speed. Afther download master doesn't have .git files. However I don't think know how to do is :(

something i have trouble understanding some of the words, because not every can translate on polish. But i ask for confidence: squash is the same as the merging?

hasufell:
> squash is the same as the merging?

Squashing means "transform multiple commits into a single one". That is done via `git rebase -i <commit-range>`. Merging preserves commits/history... any may add a new "merge commit" (unless it's fast-forward merge) on top which combines two branches/histories.

Maybe [this](https://github.com/blog/2141-squash-your-commits) illustrates it?

strugee:
@kompowiec yeah, so I want you to combine a10e813 and  0a3c868 into a single commit. That's squashing.

vyp:
I honestly don't mind merging this as is. Or otherwise @strugee can we not just click on squash and merge from the merge pull request button?

Just one thing though, entries for hotspot-shield and cyberghost should be added to en-projects-rejected.ls.

strugee:
@vyp I tend to be a little anal about [atomic commits](https://www.freshconsulting.com/atomic-commits/), which is why I rarely use Squash and Merge. It's more work, but creates a _much_ nicer history.

I don't mind doing this rebase, though. Just slipped my mind.

vyp:
@strugee Oh right, I thought you wanted all 4 commits to be 1 commit, fair enough. And same of course, but at the same time I would definitely like to see these contributions and translations added to the site, and therefore at some point I think we just have to take some sort of action (if we can and have the time) instead of waiting indefinitely or something. (Not saying you were, I see that this just slipped through :wink:)

strugee:
@vyp pushed and tested locally; please review

strugee:
Uh. Let me fix the mess I've just made :P

Issue #1471, opened 2015-12-28
Will Ubuntu/Derivatives be considered again after the 16.04 update?


1029chris:
Since Ubuntu 16.04 is disabling it's spyware by default on it's upcoming version 16.04, will Ubuntu or its derivatives be considered then?
I'd argue that Linux Mint, and the other Ubuntu derivative should be considered again after 16.04, because they provide good security through encryption. They provide an easy way for beginners to encrypt their systems. I don't think Ubuntu's packages outside of Unity contain any spyware myself. Trisquel is based on Ubuntu, and is endorsed by FSF/GNU. The FSF/GNU is arguably more concerned about free software than us, and they endorse Trisquel. Issue #642 claims that Linux Mint connects to many different domains, and it does. Because Linux Mint automatically checks for updates, so Mint connecting to so many domains would be expected.
I think this list needs a true beginners distribution. Not even Fedora is a true beginners distribution. Will the tech illiterate looking at this list try something difficult, or just keep running Windows/OSX?
Proof: http://www.pcworld.com/article/3015021/linux/ubuntu-1604-lts-wont-send-local-searches-over-the-web-by-default.html

hasufell:
> I don't think Ubuntu's packages outside of Unity contain any spyware myself.

That's your opinion. It's a question of trust though. Do you believe someone who has lied to you once?

> The FSF/GNU is arguably more concerned about free software than us

Exactly. They primarily care about license stuff, not security, so they have a completely different focus.

1029chris:
[...]

Issue #1463, opened 2015-12-21
Ubuntu 14.04 server?


ghost:
Since the data leaks and amazon ads have been removed from ubuntu, can it be considered safe for server use?

yegortimoshenko:
Merged into #1471.

Issue #1310, opened 2015-05-29
Icecat


szymon1051:
http://directory.fsf.org/wiki/GNU_IceCat

hasufell:
So they remove trademarked artwork while being considerably slower with security updates? Sound like a silly trade-off.

vyp:
> considerably slower with security updates

?

hasufell:
I was drunk yesterday, so sorry.

vyp:
Haha what. No I meant like if you have sources on how you mean, not saying you're wrong, but that I would like to know about it too. But if not, it's all good.

nodiscc:
I also would like to know more about IceCat. The page says

> The GNU Project keeps IceCat in synchronization with upstream development of Firefox while removing all trademarked artwork [..]

but the download section says

> Download version 31.6.0 (stable) released on 3 April 2015 

But the release version Mozilla Firefox is 38.x. Is this seriously maintained? (serious question) There have been quite a few changes since 31.x (aside new features, there are security and bug fixes). Did those get backported?

Are there binary releases of IceCat for non-free operating systems?

<sup>As I/other people said in https://github.com/nylira/prism-break/issues/1307, https://github.com/nylira/prism-break/issues/1235, https://github.com/nylira/prism-break/issues/1254, https://github.com/nylira/prism-break/issues/1184, https://github.com/nylira/prism-break/issues/1091, I think it would be better to provide a preconfigured/properly patched version of the latest Firefox release instead of replacing it with low-maintenance forks.</sup>

vyp:
@nodiscc Icecat targets the extended support release of Firefox, so even though the latest ESR version is 38, it was "only" released a few weeks ago, so in my opinion it's understandable that Icecat is still on 31 (the previous ESR version). But it needs a new update soon.

The question though is whether *prism break* finds that acceptable.

> Are there binary releases of IceCat for non-free operating systems?

Not that *I* know of (but I don't know very much). In fact, I'm not even sure if there are binary releases for non-*gnu linux* systems. However, I don't think that should hold it back from being listed on at least the GNU/Linux page.

vyp:
Also of note would be the "privacy protection features" listed here: http://www.gnu.org/software/gnuzilla/. It's a little bit different but if I understand correctly it's still just a GNU 'repackaging' of Firefox, much like Iceweasel by Debian.

vyp:
As of 31.5.0 (current is 31.6.0), they *seem* to support binary releases for:

[...]

PR #1683, opened 2017-02-09
Add Cloudron project

nebulade:
This contains one commit which only fixes trailing whitespace. I am not sure if this is wanted, if not, I will remove that commit.

Thanks.


strugee:
Hi, thanks for your PR. I don't recall Cloudron being discussed before; is that correct? If so we'd need to evaluate it before this can be merged.

Thanks for contributing to PRISM Break.

gramakri:
@strugee I am one of the co-founders of cloudron.io. Let me know if you have any questions. The best place to start is https://cloudron.io/get.html#selfhost. You can also see the demo at https://my-demo.cloudron.me.

nebulade:
@strugee is there any document on what the criterias for being accepted in the project list are? 

From the main Readme it looks like being an open source project is one of the main points (Cloudron is AGPLv3). Additionally from a project vision standpoint, I think Cloudron is largely aligned with your statement in the `Quality over quantity` point, focusing on allowing to stay in control of data together with providing ease of use. (see top of the page https://cloudron.io/features.html)

strugee:
@nebulade https://github.com/nylira/prism-break#project-inclusion-guidelines is the official project inclusion guidelines. However I think that list could do with a rewrite... it doesn't really make it clear to what extent we still value security. For a very clear-cut example take #1310 where we rejected Icecat because it wasn't able to keep up with upstream security patches.

I'll open an issue to formalize this and correct the list I linked to above, but broadly in terms of security we're looking for projects that:

* Understand security and take it seriously (this is very broad, but can range from "don't invent your own crypto" to "think about and design against metadata leaks")
* Offers security advantages that other projects don't (this is why e.g. Sandstorm's containerization model is particularly interesting to us)
* Can keep up with security events
* Doesn't depend on, imply or promote insecure things including and especially things in the "proprietary" column

This list is non-exhaustive, though, and we should really fix the above link.

Hillside502:
@strugee 
Cloudron is tied to
`Ubuntu 16.04 (Xenial)`
https://cloudron.io/references/selfhosting.html#installing

How does this fit with PRISM Break's Ubuntu exclusion --- such as #1463 #1471 ?

strugee:
@gramakri is Cloudron actually tied to Ubuntu? Or is it just a recommendation, not a requirement?

If it's the former, we can't list it on PRISM Break. We've rejected projects before because they require rejected dependencies (the perennial example being Signal).

gramakri:
@strugee Yes, Cloudron requires Ubuntu 16.04 Sever.

strugee:
@gramakri that's too bad. Unfortunately we can't list Cloudron in that case since we've previously rejected Ubuntu from the list.

nebulade:
That's too bad. Is the rejection still holding up, given the comments in https://github.com/nylira/prism-break/issues/1471 as well as I assume there is much difference between the desktop and the server version?

Hillside502:
@nebulade 
Looking at:-
Server and Desktop Differences - Preparing to Install - Ubuntu Server Guide
https://help.ubuntu.com/16.04/serverguide/preparing-to-install.html#intro-server-differences

"There are a few differences between the Ubuntu Server Edition and the Ubuntu Desktop Edition. It should be noted that **both editions use the same apt repositories**..."

"Ubuntu no longer has separate -server and -generic kernel flavors. These have been **merged into a single -generic kernel flavor**..."

So, please consider an additional (or replacement) OS from:-
Servers - Platforms - PRISM Break
https://prism-break.org/en/categories/servers/#operating-systems

of course, reduced to a subset as available on cloud infrastructure providers, especially --- my favourite, the Swiss-located Exoscale.

Simple cloud computing for cloud native teams | Exoscale
https://www.exoscale.ch/

How Exoscale compares to DigitalOcean and Amazon
https://www.exoscale.ch/compare/

nebulade:
I understand your position here. It is however unlikely that we will officially support any other distribution in mid term, given the additional maintenance and testing effort required. I will come back to this pull request once we do. Thanks.

strugee:
@nebulade PRISM Break has not yet made a decision regarding Ubuntu, so this rejection still stands. If we ever add Ubuntu to the list again then I'd be happy to reopen this PR.

PR #1702, opened 2017-03-06
add Nextcloud

jancborchardt:
This adds [Nextcloud](https://nextcloud.com), a completely open source fork of Owncloud.

Please review @nylira @alerque @strugee :)


jancborchardt:
@strugee thanks for the review, fixed both! :) The CONTRIBUTING.md was a bit confusing which led me to edit all the other language files too – fixed that sentence in another pull request at https://github.com/nylira/prism-break/pull/1704

source/db/en-projects.json
```diff
@@ -3918,6 +3918,117 @@
...
     "development_stage": "released",
+    "description": "A safe home for all your data",
+    "license_url": "https://github.com/nextcloud/server/blob/master/COPYING-README",
+    "logo": "nextcloud.png",
+    "notes": "Nextcloud provides a file sync & share solution that respects your privacy and is fully open source. Sync clients are available for all major platforms. Nextcloud comes with a rich ecosystem of apps including Contacts, Calendar, Video calls, Mail, Newsreader and much more.",
+    "privacy_url": "",
+    "source_url": "https://github.com/nextcloud",
+    "name": "Nextcloud",
+    "tos_url": "",
+    "url": "http://nextcloud.com",
```
strugee: Nit: `https://`

source/db/en-projects.json
```diff
@@ -3918,6 +3918,117 @@
...
+    "wikipedia_url": "https://en.wikipedia.org/wiki/Nextcloud",
+    "protocols": [
+      "CalDAV",
+      "CardDAV",
+      "IMAP",
```
strugee: We don't currently have an `IMAP` protocol. Possibly we should, but in the meantime if you can get rid of this line that'd be great :)

source/db/en-projects.json
```diff
@@ -3918,6 +3918,117 @@
...
+        "subcategories": [
+          "Enterprise Suite",
+          "File Storage & Sync",
+          "Media Publishing",
+          "Productivity",
+          "Video & Voice"
+        ]
+      },
+      {
+        "name": "Servers",
```
strugee: Probably all the GNU/Linux, BSD, etc. entries should be removed. This is consistent with other selfhosted server stuff we list (Piwik, Prosody, Postfix)

source/db/en-projects.json
jancborchardt: Fixed – woud be good to add IMAP then ;)

source/db/en-projects.json
jancborchardt: Of course, thanks, fixed! :)

source/db/en-projects.json
jancborchardt: These refer to the desktop capabilities. Much like we have apps for iOS & Android and desktop clients for Windows and macOS, they also work on Linux. :) Piwik etc does not have desktop apps, so they don’t have entries there. But we sync files, and are deeply integrated via Calendar / Contacts sync, especially in GNOME for example.

strugee: Ah, right. Forgot about that :)

PR #1706, opened 2017-03-12
Update Piwik logo

iamcryptoki:


strugee:
:thumbsup: thank you!

I wish we had an SVG version of this, but. Ah well. Merging since this is clearly an improvement.

PR #1726, opened 2017-04-07
Update Replicant links and descriptions

wiewo:
The source code was moved some time ago and the about page doesn't
exist anymore.
CyanogenMod was discontinued and the current development version is
based on LineageOS.


source/db/ar-projects.json
```diff
@@ -5097,11 +5097,11 @@
   {
     "development_stage": "released",
     "description": "توزيعة  أندرويد حرة  بالكامل.",
-    "license_url": "http://replicant.us/about/",
+    "license_url": "https://git.replicant.us/replicant/manifest/blob/replicant-6.0-dev/README.mkdn",
```
strugee: This isn't strictly a license URL, but I guess it's the best we're going to get? Probably about the same as the old link

wiewo: Yes, pretty much the same. It's beyond the scope of the project to keep a list of all subprojects and their respective licenses. The individual repos need to be checked for their licenses.

wiewo:
The updated commit now includes all 26 files.

PR #1731, opened 2017-05-05
Update OpenNIC-related URLs

Fusl:


alerque:
Thanks for the contribution @Fusl!

Issue #1703, opened 2017-03-07
Adding Subgraph OS in the operating systems section


ghost:
https://subgraph.com

Still in alpha so it should have the "Experimental" label.

hasufell:
Interesting read (don't consider that a show-stopper though): https://micahflee.com/2017/04/breaking-the-security-model-of-subgraph-os/

I wasn't able to find the full license of the project and got no answer in the IRC channel yet. 

The post author from that blog also says

> The Subgraph TemplateVM project is basically stalled right now because it appears that Subgraph isn’t 100% open source yet, and non-Subgraph devs don’t have the ability to build their own ISOs yet.

If that's true, then we should probably wait.

hasufell:
OFTC network, `#subgraph` channel

```
[13:30] <hasufell> where is the license of subgraph documented?
[13:30] <hasufell> I see GPLv3+ on wikipedia but no reference
[13:31] <hasufell> e.g. oz is BSD-3
[13:57] <xmux> Since subgraph os is a distribution of various software by various authors it isn't distributed under a single license
[14:03] <hasufell> xmux: why not? Licenses can be compatible. You can license a project GPL-3 even if it contains BSD-3 parts.
[14:04] <hasufell> if it's not a single license, then why does wikipedia say GPLv3+? If it's a set of licenses, then where are they documented?
[14:05] <xmux> I didn't write the wikipedia page, that information sounds not exactly correct
[14:07] <xmux> licenses are documented in /usr/share/doc/*/copyright
[14:08] <hasufell> erm?
[14:08] <hasufell> those are package licenses
[14:08] <hasufell> I am talking about subgraph os
[14:10] <xmux> Since we can't relicense other people's software, those are the licenses that matter. It doesn't make sense to ask what single license the entire project is distributed under
[14:10] <hasufell> sure it does, I'm talking about the subgraph os bits, not about the packages you redistribute
[14:10] <hasufell> e.g. oz, PM maybe, layouts, scripts...
[14:11] <hasufell> basically every distro has a license for their tools and base files
[14:11] <xmux> all of the projects have LICENSE files on github
[14:11] <xmux> generally, we choose BSD
[14:12] <hasufell> https://github.com/subgraph/defector don't see a license there
[14:13] <xmux> That's an abandoned experiment, I think.
[14:13] <xmux> but yes it should have a license
[14:13] <hasufell> https://github.com/subgraph/subgraph-os-apparmor-profiles and here
[...]

PR #1738, opened 2017-05-22
Add Subgraph OS wrt #1703

hasufell:


hasufell:
Hello

hasufell:
Can someone explain to me why this PR is being ignored for almost a year now?

alerque:
I realize you called out the issue in the title, but since Github doesn't link that, having a PR description with a link would help. For reference the discussion for this issue is at #1703 ... current status is the issue was tagged as blocked pending source releases. I would suggest finishing up the discussion including @strugee's points over there in the issue instead of here in the PR.

hasufell:
> current status is the issue was tagged as blocked pending source releases.

The sources are released. There was only FUD in the thread.



> Qubes is probably better

I don't see how that is relevant. Qubes is still listed. SubgraphOS has a clear security/privacy focus. Way more than debian.

strugee:
@hasufell responded to in https://github.com/nylira/prism-break/issues/1703#issuecomment-363870815 to keep things centralized.

alerque:
@strugee I've had a couple reads through the comment snafu over licensing now as well as reviewing where the project is at (not super in depth, but generally poking around) and this has a 👍 for merge from me, for whatever that's worth.

strugee:
LGTM although I think we should add the Experimental label? subgraph.com says it's alpha

alerque:
Yes, an experimental tag until upstream starts branding their releases as something other than alpha is probably a good move. I don't always put a lot of stock in that, some projects tag stable releases long before they  should be considered so and others wait years after they are rock solid to drop the label (looking at you KeepassX), but in this case at least for a while that sounds fair.

hasufell:
https://github.com/nylira/prism-break/pull/1738/files#diff-1d65ca0b1886a69e5ab16eaaa239e441R3 is this not the experimental tag?

strugee:
@hasufell wow, I feel very dumb. Nice catch, LGTM.

PR #1762, opened 2017-07-21
Rename "The Tor Browser Bundle" to "Tor Browser"

### LINKED ISSUE #1756: The Tor Browser Bundle name is deprecated, it's only Tor Browser
The Tor Browser Bundle name is deprecated, they used that name when they bundled other things such as Privoxy.

vyp:
Closes #1756.


vyp:
Woops, forgot about CONTRIBUTING.md, don't merge yet.

vyp:
Ah I thought about that, but assumed the inheritance would only work if the whole project was missing. But `description` keys are translated so I thought best not to do that, to keep those translations. So can I just remove for example the duplicate `note` keys?

strugee:
@vyp, yeah, just remove the duplicated English keys and any translations that are now out-of-date (that's what I've been telling people, since it seems like we want to "reset" those translations to get new versions, right?)

strugee:
And to clarify further, this:

> So can I just remove for example the duplicate `note` keys?

is spot-on.

vyp:
Done, is this what you meant? I added it as another commit in case I misinterpreted, so wouldn't lose previous work. But of course feel free to squash it if you merge!

PR #1764, opened 2017-07-21
Remove owncloud

vyp:


source/db/en-projects.json
```diff
@@ -3655,7 +3655,7 @@
     "description": "A safe home for all your data",
     "license_url": "https://github.com/nextcloud/server/blob/master/COPYING-README",
     "logo": "nextcloud.png",
-    "notes": "Nextcloud provides a file sync & share solution that respects your privacy and is fully open source. Sync clients are available for all major platforms. Nextcloud comes with a rich ecosystem of apps including Contacts, Calendar, Video calls, Mail, Newsreader and much more.",
+    "notes": "Nextcloud provides a file sync & share solution that respects your privacy and is fully open source. Sync clients are available for all major platforms. Nextcloud comes with a rich ecosystem of apps including Contacts, Calendar, Video calls, Mail, Newsreader and much more.\n\nNextcloud originated as a fork of ownCloud. It is preferred because ownCloud has an \"open core\" product model and lacks some security features Nextcloud has.",
```
strugee: I'm on mobile so I can't check, but isn't this interpolated directly into the HTML? So I _think_ these `\n`s should be `<br />`s or `</p><p>` or something.

vyp: I'm not 100% but I'm pretty sure it's markdown. Although inline HTML is allowed in markdown so maybe that could work too. But this is consistent with other project's notes and descriptions.

strugee: @vyp yep, seems to work fine. Dunno what I was thinking :P

PR #1772, opened 2017-07-27
Add `lint` Make target to lint JSON

strugee:
Not sure if this is actually the best way to do this since as I've said before, I am not very good at Makefiles :P

It does work, however.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ init: node_modules ;
 test: en ;
 
 lint:
-	find source -type f -name '*.json' -exec jsonlint -q '{}' \;
+	find source -type f -name '*.json' -print -exec jsonlint -q '{}' \;
 
 # Start fresh and rebuild everything
 reset: | clean default ;
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ init: node_modules ;
 test: en ;
 
 lint:
-	find source -name '*.json' | xargs -n 1 jsonlint -q
+	find source -type f -name '*.json' | xargs -n 1 jsonlint -q
 
 # Start fresh and rebuild everything
 reset: | clean default ;
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ LANGUAGES := $(notdir $(basename $(wildcard source/locales/*.json)))
 ASSETS := $(notdir $(wildcard source/assets/*))
 
 # Mark all rules that don’t actually check whether they need building
-.PHONY: default test init reset all $(LANGUAGES) assets css html html_% clean watch watch_css sync localize_%
+.PHONY: default test lint init reset all $(LANGUAGES) assets css html html_% clean watch watch_css sync localize_%
 
 # Turn on expansion so we can reference target patterns in our dependencies list
 .SECONDEXPANSION:
@@ -53,6 +53,9 @@ init: node_modules ;
 
 # Use building English language as a check to see if everything works
 test: en ;
+
+lint:
+	find source -name '*.json' | xargs -n 1 jsonlint -q
 
 # Start fresh and rebuild everything
 reset: | clean default ;
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ endif
 #----------------------------------------------------------------------
 
 # Explicitly set the default target to do everything that isn’t already done
-default: | init assets all public ;
+default: | init lint assets all public ;
 
 # Run anything that needs doing post-checkout to make this buildable
 init: node_modules ;
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ init: node_modules ;
 test: en ;
 
 lint:
-	find source -type f -name '*.json' | xargs -n 1 jsonlint -q
+	find source -type f -name '*.json' -exec jsonlint -q '{}' \;
 
 # Start fresh and rebuild everything
 reset: | clean default ;
--- a/package.json
+++ b/package.json
@@ -19,7 +19,9 @@
     "stylus": "~0.54.5",
     "uglify-js": "^2.7.5"
   },
-  "devDependencies": {},
+  "devDependencies": {
+    "jsonlint": "^1.6.2"
+  },
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1"
   },
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,6 @@ node_js:
     - "4"
     - "6"
     - "8"
-script: "make test"
+script:
+  - "make lint"
+  - "make test"
--- a/source/locales/zh-TW.json
+++ b/source/locales/zh-TW.json
@@ -49,7 +49,7 @@
   "XMPP Servers": "XMPP 伺服器",
   "warning-android": "警告: 提供給手機及平板電腦的Android作業系統常被修改、加入來自Google或其他來源的私有軟體而可能對你的隱私造成危害。我們強烈建議替換成[Replicant](../../projects/replicant) (或者爲你的裝置編譯Android源碼).<br><br>推介閱讀: [Mission Impossible: Hardening Android for Security and Privacy](https://blog.torproject.org/blog/mission-impossible-hardening-android-security-and-privacy), by Mike Perry at The Tor Blog.",
   "warning-ios": "警告: 蘋果 iOS 裝置受稜鏡的影響。 使用 iOS  會洩露你的隱私。我們強烈建議你使用 Anrdoid 裝置代替蘋果的 iOS 裝置。 [Replicant](../../projects/replicant).",
-  "warning-mac": "警告: 蘋果 macOS 系統受稜鏡的影響。使用 macOS 會洩露你的隱私。我們強烈建議你使用 [Linux](../gnu-linux) 或 [BSD](../bsd) 作業系統替換 macOS"。,
+  "warning-mac": "警告: 蘋果 macOS 系統受稜鏡的影響。使用 macOS 會洩露你的隱私。我們強烈建議你使用 [Linux](../gnu-linux) 或 [BSD](../bsd) 作業系統替換 macOS。",
   "warning-windows": "警告: 微軟視窗系統受稜鏡的影響。使用微軟視窗 會洩露你的隱私。我們強烈建議你使用 [Linux](../gnu-linux) 或 [BSD](../bsd) 作業系統替換微軟視窗。",
   "Communication": "通訊",
   "Anonymity": "匿名性",


strugee:
@alerque care to review?

strugee:
Just pushed a fix for a syntax error that this caught - although it didn't seem to be a problem for the automatic builds? I dunno. Anyway...

Makefile
```diff
@@ -54,6 +54,9 @@ init: node_modules ;
 # Use building English language as a check to see if everything works
 test: en ;
 
+lint:
+	find source -name '*.json' | xargs -n 1 jsonlint -q
```
vyp: Maybe there should also be a `-type f`?

strugee: @vyp nice catch; just pushed a fix. I don't think it changes anything but it's always nice to be correct :)

Makefile
```diff
@@ -54,6 +54,9 @@ init: node_modules ;
 # Use building English language as a check to see if everything works
 test: en ;
 
+lint:
+	find source -type f -name '*.json' | xargs -n 1 jsonlint -q
```
hasufell: Don't pipe the output of `find`, use `-exec` or `-execdir`, something like

```sh
find source -type f -name '*.json' -exec 'jsonlint -q {}' +
```

strugee: _Wow_. Completely forgot `find -exec` was a thing. Derp.

Just pushed a fix.

package.json
```diff
@@ -19,7 +19,9 @@
     "stylus": "~0.54.5",
     "uglify-js": "^2.7.5"
   },
-  "devDependencies": {},
+  "devDependencies": {
+    "jsonlint": "^1.6.2"
```
hasufell: why `^`? Nothing will really "break" on `>=`.

Makefile
```diff
@@ -54,6 +54,9 @@ init: node_modules ;
 # Use building English language as a check to see if everything works
 test: en ;
 
+lint:
```
hasufell: Should this be part of the `default` target? People will probably forget to run it explicitly

package.json
strugee: Cause it's what `npm install --save-dev` generated. It seems better to me to just play it safe?

Makefile
strugee: Good question. That makes a lot of sense to me although note that CI already runs it.

package.json
hasufell: How is that "safe"? Is there any reason we don't always want the __latest__ version of `jsonlint`? I can't see any.

strugee: If they release a semver-major version that changes the CLI interface, this won't pull in that version. But with `>=` it would and we would break.

hasufell: You mean in the unlikely case `-q` is removed? When that happens, you just fix it. This is not part of a dependency that may silently introduce runtime behavior change of the application.

strugee: But before it's fixed contributors will see unexpected failures and possibly not see why, making for a worse contributor experience. It's not a huge deal but I just don't see why we would opt into that on purpose when there's a really easy way to avoid it.

vyp: I'm with @strugee, keeping things running smoothly is most important. :+1:

hasufell: > But before it's fixed contributors will see unexpected failures

That's an assumption. These can also happen in micro bumps, when the interface is not changed, but simply linter rules are added in a backwards compatible manner.

When using `^` I am sure you will never look back and check for new major versions. No one does, forcing us to stick with ancient versions. I don't see why anyone would not want the __very__ latest linter ruleset. `>=` does exactly that.

vyp: > When using ^ I am sure you will never look back and check for new major versions.

I mean this is also an assumption in part. I don't see why can't just check for new versions ourselves. I feel like a contributor's (potential) experience is more important than our convenience. I see what you mean, not having the latest ruleset could also result in a bad experience for the contributor, but again, we can have the best of both worlds at the cost of some imo very minor convenience. :v:

strugee: > These can also happen in micro bumps, when the interface is not changed, but simply linter rules are added in a backwards compatible manner.

That would not be backwards-compatible. I'm wondering if you've actually read the [semantic versioning spec][spec]?

I mean honestly if this is _really_ a concern, we can enable [Greenkeeper](https://greenkeeper.io/). That's probably a good idea anyway, honestly. This whole argument may be mostly irrelevant too since upstream looks [pretty inactive][upstream]. I looked for a better alternative on npm but didn't really find any.

 [spec]: http://semver.org/
 [greenkeeper]: https://greenkeeper.io/
 [upstream]: https://github.com/zaach/jsonlint

Makefile
strugee: @vyp what do you think of this?

strugee: Pushed a change addressing this

vyp: Yeah this is a good suggestion. :+1:

package.json
hasufell: > That would not be backwards-compatible. I'm wondering if you've actually read the semantic versioning spec?

Yes, I use it for my rust and haskell libraries. And I consider added linter rules as backwards-compatible feature additions, so the minor version will be bumped, not the major one.

alerque:
@strugee Sorry a bit behind on Github notifications. Having a look now.

package.json
alerque: @hasufell That sounds like something I wouldn't expect to find in a minor dot release. Adding new lint rules could easily be _not_ backwards compatible for some projects. In any event it's neither here nor there...

I was going to suggest Greenkeeper when I first saw this thread before I got down to where it was mentioned. That's the solution for "always having the very latest", not the changing the requirements. In this case I'd suggest keeping what `--save-dev` defaulted too.

Makefile
alerque: I would actually suggest _not_ having as part of the default run. I like to keep the default target as lean as passible, sort of a minimum run to get everything in place of of the box. Extra things you'd use during a development cycle but that might not be needed in all cases shouldn't be part of this.

Note for example we're not running `test` as part of `default`. It should of course be run after making code changes but you might not want it at every step as you work through something. Having `lint` as part of the `ci` target is also a good thing.

An `all` target separate from `default` could include a heavier load.

Makefile
```diff
@@ -46,14 +46,17 @@ endif
...
+default: | init lint assets all public ;
 
 # Run anything that needs doing post-checkout to make this buildable
 init: node_modules ;
 
 # Use building English language as a check to see if everything works
 test: en ;
 
+lint:
+	find source -type f -name '*.json' -print -exec jsonlint -q '{}' \;
```
alerque: This probably doesn't need `-print` (does `jsonlint` not say what it's working on anyway?), and I'd always reach for `-execdir` unless the linter specifically can't handle that. Probably not a big deal in this case but it should be the default unless there is a reason it needs to run from the project root.

Makefile
vyp: Yes actually I think @alerque makes a good point here, agreed. But the `default` target already contains `all` and everything else though, so in a way the `default` target is already acting as some sort of 'all'. So I think it's actually fine the way it is.. and `make test` is what @alerque is looking for? @alerque could you confirm?

Makefile
hasufell: First of all, this shouldn't have `\;`, but `+`, second... `-exec` is more efficient than `-execdir` here, because it will cause only one invocation of `jsonlint`.

package.json
hasufell: > Adding new lint rules could easily be not backwards compatible for some projects.

Again: I consider that backwards compatible. Whether or not your code passes the linter is not part of the public API. Even micro bumps can break that (when they fix broken rules), so...

> I was going to suggest Greenkeeper

It's good to avoid upper bounds, when they are not strictly necessary (and they are often not for build tools).

strugee: > Again: I consider that backwards compatible. Whether or not your code passes the linter is not part of the public API. Even micro bumps can break that (when they fix broken rules), so...

Then you disagree with the vast majority of the npm community, who think that if it breaks _anything_, including your test suite, then it's a semver-major change. Adding new rules may not affect the API _interface_ but it certainly impacts API _behavior_. That makes it backwards-incompatible.

strugee:
> @strugee Sorry a bit behind on Github notifications. Having a look now.

Hah, no worries :D

I'm _always_ behind on GitHub notifications.

Makefile
strugee: > does `jsonlint` not say what it's working on anyway?

It doesn't and even worse, it doesn't report the filename along with errors :'(

Re: `-execdir`, I tried that on my machine and `find` gave me some error having to do with my personal `$PATH`. AFAICT from `find(1)` the problems with `-exec` aren't a big deal here, so I just didn't bother to debug and just took the easier path. Happy to be told otherwise though.

> EDIT: strike that, `jsonlint` only takes one filename anyway it seems

Yep :/

Makefile
strugee: Yeah, `default` IIRC builds everything. Linting is relatively cheap in terms of time, so I thought it would be better to run `lint` in the default target because _if_ the contributor introduced an error, `make` would fail much, much faster.

Maybe we should have `default` just build the English version though, and introduce an `all` target similar to the current `default`?

Makefile
alerque: > It doesn't and even worse, it doesn't report the filename along with errors :'(

I'm thinking of a different linter. So yes, you're quite right to add a print so one might be able to puzzle out where it died in the event of failure.

> the problems with -exec aren't a big deal here

Ya I gave a :+1: for this changeset as is because it's pretty minor, more a note for future reference. The path error is actually something you should think about fixing in your RC files. Find is kind of anal about that, but it does have a point about security and working with project code you download off the internet it's something to think about.

Makefile
alerque: > introduce an all target similar to the current default

For some reason I thought that's the way it was, but I see that's not quite the case. Yes that is what I'd recommend too: `default` being the bare bones minimum to go from zero to working build, and `all` being a sort of kitchen sink build. Meanwhile `test` should probably call `lint` as a prerequisite:

    test: lint en ;

vyp: :+1: on this. @nylira how do you feel about it? (Since you're the one that ends up building and publishing all the changes everytime for the actual site :wink:)

package.json
hasufell: > who think that if it breaks anything

Doesn't seem to be particularly true for jsonlint, unless you are saying the exit code is not part of the API behavior: https://github.com/zaach/jsonlint/commit/b9a537a3c1c4c475b9be09502680ab5c95039edb

Makefile
hasufell: > AFAICT from find(1) the problems with -exec aren't a big deal here

Neither the [filename attack](https://www.gnu.org/software/findutils/manual/html_mono/find.html#Problems-with-_002dexec-and-filenames), nor the symbolic link dereferencing [race condition](https://www.gnu.org/software/findutils/manual/html_mono/find.html#Race-Conditions-with-_002dexec) are relevant here afais, so `-exec` is indeed fine.

strugee: @alerque right, that's true :)

I'll try to do that then.

strugee:
Merging and will file a followup for https://github.com/nylira/prism-break/pull/1772#discussion_r130235663.

PR #1774, opened 2017-07-27
Update rejected projects

strixaluco:


source/db/en-projects-rejected.ls
```diff
@@ -290,14 +290,14 @@ projects-rejected =
     name: "Facebook"
     categories: [
       * name: "Web Services"
-        subcategories: ["Email Accounts", "Instant Messaging", "Social Networks", "Video & Voice"]
+        subcategories: ["Email Accounts", "Social Networks", "Video & Voice"]
       * name: "Servers"
-        subcategories: ["Email Server", "Instant Messaging", "Social Networks", "Video & Voice"]
```
vyp: Should probably keep this because one can still do instant messaging on facebook.com.

strugee: I think the intention was to have "Facebook Messenger" appear in the instant messaging category _instead of_ Facebook?

strixaluco: I apologize for any confusion.
Yes, the idea was to add Facebook Messenger to "Instant Messaging" instead of Facebook

vyp: Yeah I know but you can still use the small chat boxes on facebook.com instead of having to use messenger. Would it too confusing/too much clutter to have both under instant messaging you guys think?

strixaluco: I don't have any strong opposition against having them both.

vyp: Vice versa, up to you @strugee.

vyp: I guess it would less confusing to have only faceboook-messenger anyway. :+1:

strugee: Ah sorry @vyp, thanks for merging this. Yes, I think it's better to err on the side of having less projects on the page, and I don't think anyone will really pay too much attention to the details we're discussing here :thumbsup:

PR #1779, opened 2017-07-30
Kompowiec's changes

strugee:
Somehow botched the force-push on #1665, so I'm just resubmitting. @vyp if you could review this? Basically just the changes you mentioned in https://github.com/nylira/prism-break/pull/1665#issuecomment-318642971.


strugee:
> since they got wiped out

By this PR?

vyp:
Oh no I meant that they got wiped out in #1665 so couldn't confirm if these are the same as those. Anyway, I think it should be fine to merge now. @kompowiec Let us know if there were any mistakes here. And again, thank you!

kompowiec:
I think so is good, doesn't seen mistakes.
Also thank you, You did it better than me.

PR #1790, opened 2017-08-17
Switch to Contributor Covenant

### LINKED ISSUE #1657: Open Code of Conduct is dead
Our Code of Conduct isn't being worked on anymore: http://todogroup.org/blog/followup-open-code-of-conduct/

In particular, http://todogroup.org/opencodeofconduct/#PRISM-Break/abuse@prism-break.org no longer shows our project name and abuse contact for me, though it could just be a browser misconfiguration on my end.

Not sure if we want to do anything about this, but I thought I'd bring it up for discussion anyway.

strugee:
Fixes #1657


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
```
hasufell: edit comments? I don't think we should allow that.

strugee: @hasufell I've used this in the past, when a troll was posting IPs and stuff which were presumably doxxed. I think GitHub makes it obvious enough when comments are edited by maintainers for this to be a non-issue.

hasufell: I cannot follow that reasoning. In that case you should just delete the post, not edit it.

alerque:
Where did the "covenant" wording come into play here? I haven't been following super closely but if the choice is local to this project (not matching some external project) I would suggest different diction.

vyp:
> I would suggest different diction.

Why? As in, what's wrong with covenant?

It's from https://www.contributor-covenant.org/.

CODE_OF_CONDUCT.md
vyp: @hasufell One instance where this is useful is the original post of an issue/PR itself, which cannot be deleted. (Unless in some cases if you contact GitHub about it, I think. But it's more convenient to just edit things out. And, for example, especially in cases of doxxing, you'd want to remove things as quickly as possible with your own agency without having to wait and be dependent on GitHub.)

strugee: @hasufell can you comment on https://github.com/nylira/prism-break/pull/1790#discussion_r147551520? I don't really see the problem especially since, again, GitHub clearly shows the edit history for a post.

strugee:
Self-assigning so this doesn't drop off my radar again.

yegortimoshenko:
I'm not comfortable with CoCs because they place the project (usually, original maintainer) into unfair position of power (to judge others), which is against decentralized nature of Git.

They also are rather legalistic and imply hierarchical centralized structure which maintains who can or can not contribute to the project.

Also, I'm not sure which partciular problem does it solve: if someone behaves improperly, we don't need a document to ask them to stop, or lock them from participation. Examples of offenses that are generally considered bannable but are not in Contributor Covenant include spam, making malicious changes, or probably even having different presuppositions (i.e. not believing that privacy is important and participating in the project by repeatedly commenting about that).

Also, most of the stuff listed in this code of conduct is highly subjective and the same, say, comment can be interpreted as either offensive or friendly depending on the person reading it.

In short, my point is that it would be nicer to embrace the fact that we ban and remove comments when we think it's right, without trying to make it look like there is an objective reason for us to do that.

To ensure that there is no misunderstanding, I should state that I'm all for treating people well. 

And, by all means, I don't think that codes of conduct are effectively harmful, because I don't believe that people read and honor them. I think, changing one with the other won't effectively change anything, so feel free to merge.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at peng@nylira.com. All
+complaints will be reviewed and investigated and will result in a response that
+is deemed necessary and appropriate to the circumstances. The project team is
+obligated to maintain confidentiality with regard to the reporter of an incident.
```
yegortimoshenko: I think this is the only place where the old code of conduct is qualitatively different from this one. Open Code of Conduct states: 

> We will respect confidentiality requests for the purpose of protecting victims of abuse.

Contributor Covenant enforces reporter confidentiality (i.e. no other option). In most cases, it's good, but it might be prone to abuse (there might be someone who provably falsely reports people on regular basis, and it would be nice to be able to publish that information). Respecting, but not enforcing such requests seem to strike a balance better.

yegortimoshenko:
Merging, because it has been agreed upon (#1657 is labeled as "help wanted"), and also it seems that it's better in some regards (Open CoC has infamous "hug without consent" clause).

I still think that this is not reasonable, I would very much prefer a [Debian-style](https://www.debian.org/code_of_conduct) "be excellent to each other" code of conduct, preferably written by us, instead of a legalistic passive-agressive boilerplate (sorry for being grumpy today :-).

PR #1795, opened 2017-08-19
Remove Ostel

### LINKED ISSUE #1793: Ostel.co is dead. Delete it.
Ostel.co is dead. Service is officially closed. It's time to delete it from list (SIP server).
https://guardianproject.info/2017/08/10/ostel-co-is-permanently-offline/

vyp:
The general note about encryption only being supported for video and voice but not for instant messaging (e15781f) is not true anymore for most of the clients I believe. It originally only applied to Linphone (#1302) and CSipSimple (https://github.com/nylira/prism-break/pull/1316#issuecomment-108223563), but CSipSimple is not on PRISM Break anymore, so I also rewrote that part to only apply to Linphone. Hope that's okay, otherwise I can put it into a separate commit or PR or something.

There are still some references to Ostel (just do `git grep -i ostel`) in the translations, but I'm not exactly sure how to remedy them unfortunately. 🤷‍♀️

Edit: Closes #1793.


mimi89999:
I think the *Notes* from voice and video should be removed as they are outdated and therefore can lead to misinformation...

vyp:
@mimi89999 Does Linphone now support encrypted instant messages?

mimi89999:
@vyp I don't know. Probably not. I haven't realized that that note is only about Linphone and not the entire section...

mimi89999:
I will remove all notes under Video & Voice in https://github.com/nylira/prism-break/pull/1797

vyp:
Yeah so once this is merged, the notes section for video and voice (for English) will only be talking about Linphone.

mimi89999:
What is blocking this PR from being merged?

vyp:
@mimi89999 just that I guess no one got the time to review it. I think it's good form to make PRs when someone wants code in the default (master) branch, instead of pushing directly to it, even if only to at least let other maintainers know about changes coming in. (I'll only push directly if I feel there are some corrections to existing code that should be there immediately, or something like that.)

With that being said though, I'll merge this now! :grin:

PR #1797, opened 2017-08-19
Remove all Video & Voice notes

mimi89999:


source/locales/en.json
```diff
@@ -138,7 +138,7 @@
...
   "irc-notes": "--",
   "Video and VoIP": "Video conferencing &amp; VoIP",
-  "video-voice-notes": "<p>Get a free Sip-account for <strong>Jitsi</strong> and/or <strong>CSipSimple</strong> with The Guardian Project’s <a href='#sip-servers'>Ostel</a> service, or for <strong>Linphone</strong> with the <a href='http://www.linphone.org/free-sip-service.html'>Linphone Free SIP service</a>.</p><p><strong>Note:</strong> Encryption is only supported for voice and video, but not for instant messaging.</p>",
```
vyp: This change will conflict with #1795, so just keep this line untouched here.

vyp:
Other than the one comment, I think this is probably reasonable. Just unfortunate that they cannot be fully up to date, but oh well. I'll leave this open for a while though in case anyone thinks otherwise.

mimi89999:
@vyp Since this is just removing old translations, include that in your other PR. Just run `sed -i -e 's/video-voice-notes.*/video-voice-notes\": \"--\",/' *.json` in `source/locales/`

vyp:
I'm happy to do that, but if you'd care to reopen this, I still prefer to have this in a separate issue/PR like this since this is more of a change than just 'removing ostel', which is what that PR should mainly be about.

Or is it that you don't have the time to make the requested change? (Which is fair.)

mimi89999:
@vyp Done

vyp:
@mimi89999 Thanks! :)

PR #1799, opened 2017-08-20
Remove Thunderbird from XMPP

### LINKED ISSUE #1755: Remove Thunderbird from XMPP?
From https://github.com/nylira/prism-break/issues/1688#issuecomment-308356717:

> I recommend removing Thunderbird. It doesn't support SRV records, causing problems for many larger servers.

If this is true we should probably remove Thunderbird from the XMPP category.

mimi89999:
Closes #1755

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -3460,8 +3460,7 @@
       "RSS",
       "S/MIME",
       "SSL/TLS",
-      "Tor",
-      "XMPP"
+      "Tor"
     ],
     "categories": [
       {
--- a/source/db/ca-projects.json
+++ b/source/db/ca-projects.json
@@ -3460,8 +3460,7 @@
       "RSS",
       "S/MIME",
       "SSL/TLS",
-      "Tor",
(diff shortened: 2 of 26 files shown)


mimi89999:
@vyp Could you review?

source/db/ar-projects.json
```diff
@@ -3461,7 +3461,6 @@
       "S/MIME",
       "SSL/TLS",
       "Tor",
```
vyp: The trailing comma for the last element has to be removed otherwise it's not valid json. (And same goes for the rest of the files.)

mimi89999: @vyp Sorry. I missed that... Fixed with `sed -i -e ':a;N;$!ba;s/Tor\",\n\ \ \ \ ],/Tor\"\n\ \ \ \ ],/' *-projects.json`.

vyp:
LGTM now, thanks @mimi89999! :smiley:

mimi89999:
Thanks

Issue #1709, opened 2017-03-17
Clean up OS section


hasufell:
IMO, the OS section is overloaded which makes it harder to recommend more interesting approaches that are actually security relevant like https://github.com/nylira/prism-break/issues/1703

I'd suggest to:
1. remove all esoteric occurences that are hard to maintain or just not fit for regular users
2. focus on large distros, not small forks that play around with a few packages and licenses
3. keep one big rpm-distro and one deb-distro and one *BSD

For example, __remove__ the following:
* Fedora and CentOS (keep OpenSUSE instead)
* gNewSense (it's debian anyway)
* OpenBSD and NetBSD (not fit for regular users, FreeBSD is more appropriate there)
* DragonFly BSD (we have FreeBSD already)
* Parabola and ArchLinux (I don't see how those are security relevant and they're not one of the really big distros)
* Gentoo (big and active, but not fit for regular users)
* Scientific Linux (small distro afaik, security relevance?)

Then there are weird things like SME Server and Yunohost. Misconfigured (aka preconfigured) servers should probably not be recommended?

I also feel server-specific stuff like AntiPrism and ArkOS are a little bit misplaced. I'd rather add another section _Operating Systems (Servers)_ like for live and mobile systems.

strugee:
In general I'm very much :thumbsup: on this. Some specific questions follow.

> Fedora and CentOS (keep OpenSUSE instead)

AFAIK Fedora has a stronger commitment to software freedom than OpenSUSE, though admittedly I haven't checked in a long while. Is that not correct?

> gNewSense (it's debian anyway)

I agree it doesn't make sense to list both gNewSense _and_ Debian, but I'm not sure how I feel about leaving Debian as-is since we don't have a note or anything about disabling contrib and non-free

> OpenBSD and NetBSD (not fit for regular users, FreeBSD is more appropriate there)

OpenBSD seems to have _some_ value as an operating system for people who need security that goes above and beyond the rest of the list, so I vote to keep that. NetBSD and DragonflyBSD don't seem to be there for any particular reason, though.

> Then there are weird things like SME Server and Yunohost. Misconfigured (aka preconfigured) servers should probably not be recommended?

This probably merits a longer discussion - can you file a new issue?

:thumbsup: on the "server OS" bit. It might make sense to move CentOS in that section too, and add Debian, as more "traditional" server systems (as opposed to "managed" systems like ArkOS).

hasufell:
[...]

PR #1800, opened 2017-08-21
Remove debian comment about disabling contrib and non-free repositories

vyp:
See https://github.com/nylira/prism-break/issues/1709#issuecomment-323799573


mimi89999:
One thing you maybe forgot about:
If the installer detects hardware that won't work with free software (usually Wi-Fi cards), it will give an option and suggest to enable non-free and contrib and to install the driver/firmware from there....

hasufell:
Hm, I think we are in nitpicking land anyway, but if the installer already gives a warning, is that not sufficient? And if it isn't, shouldn't we suggest a fix for the debian installer (e.g. clarifying the implications) instead of adding that info to prism-break notes?

vyp:
Well @mimi89999 said it suggests it as an option, not necessarily as a warning, to be nitpicky. And I would agree that if that's the case, maybe we could have a note warning about using that option? But if it is already a 'warning' then I don't think any further note is needed. (And I'm not sure what it is myself because I haven't installed debian in years.)

strugee:
Did some investigation, looks like these strings are defined here: https://anonscm.debian.org/cgit/d-i/apt-setup.git/tree/debian/po/templates.pot#n471. IIRC "use {non-free,contrib} software" is the title of the message box, which has the paragraph, two newlines and the question. Then button for yes/no. Again, IIRC, no is selected by default.

hasufell:
I think that's enough for a warning. No need for handholding.

strugee:
@vyp can you rebase?

strugee:
Also while you're at it maybe shorten the commit message so it's [under 50 characters](https://chris.beams.io/posts/git-commit/#limit-50); maybe

>Remove advice on disabling Debian contrib/non-free repos

would work?

hasufell:
> under 50 characters;

🙄

strugee:
@hasufell hm? I like a nice commit log; if it's not under 50 that's fine but it's nice to shoot for.

hasufell:
That's arbitrary preference (and most projects I know and have contributed to have a different preference). We shouldn't bother contributors with that. It's useless.

strugee:
@vyp ping re: https://github.com/nylira/prism-break/pull/1800#issuecomment-361105758

yegortimoshenko:
@strugee: I've rebased it on top of master.

yegortimoshenko:
Thanks!

PR #1806, opened 2017-09-03
Update pfSense logo

strixaluco:


strixaluco:
Sorry, pulled another request with Gogs logo

PR #1811, opened 2017-09-08
Make 'all' target and other cleanup

### LINKED ISSUE #1781: Make a `make all` target
Per https://github.com/nylira/prism-break/pull/1772#discussion_r130235663 we should make the current Make `default` task do less and introduce an `all` target that does everything.

/cc @alerque

alerque:
Closes #1781 as well as fixes a few other bugs in the make process (including things that appeared to work but only accidentally and that should have worked if make didn't have some wonky bugs).

Please test this before merging. I ran a couple quick tests on my machine but certainly didn't cover every possible combination. Notably `make -jN` (where N is something like the number of CPU cores you have) should actually work now to build multiple languages at once. It sometimes worked before but only _after_ at least one try (or if by chance the first try randomly got one job done before the second one failed!).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,8 +87,9 @@ css: init public/assets/css/screen.css
 .PHONY: html
 html: init $(foreach LANGUAGE,$(LANGUAGES),html_$(LANGUAGE))
 
-.PHONY: html_%
-html_%: public/%/index.html
+HTMLLANGS = $(foreach LANGUAGE,$(LANGUAGES),html_$(LANGUAGE))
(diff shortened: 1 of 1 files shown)


alerque:
@nylira Why is `yarn.lock` stored in the repository (see e7562808c)? The specifications as far as what versions we want of what are in `packages.json` and the lock file describes the state of your system in relation to that, not of the repository. As such I don't think it should be tracked (unless there is something I'm missing here).

nylira:
`yarn.lock` is no longer needed by npm5, will remove.

PR #1814, opened 2017-09-25
complete the empty shields proprietary

kompowiec:
I study previous commits so i hope that everything will work.


yegortimoshenko:
I don't think this should be merged because we only list notable proprietary software, and most of the software that this pull request adds does not meet that criterion in my opinion.

kompowiec:
tbh i used alternativeto.net to fill empty table with proprietary software and in fact, not all of these services are popular eg. Cell Spy Cather (Anti Spy) but rest IMO is popular - hamachi, EC2, azure...

yegortimoshenko:
All PNG files contain this text:

```
+--2017-09-25 09:19:58--  https://d2.alternativeto.net/dist/icons/zent-io_60086.png
+Translacja d2.alternativeto.net (d2.alternativeto.net)... 137.117.84.54
+Łączenie się z d2.alternativeto.net (d2.alternativeto.net)|137.117.84.54|:443... połączono.
+Żądanie HTTP wysłano, oczekiwanie na odpowiedź... 404 Not Found
+2017-09-25 09:19:59 BŁĄD 404: Not Found.
```

Hamachi might be, and perhaps Threema. Others not so much. I would support only including the most popular proprietary stuff, because otherwise there would be no end to the list. Either way, if you're interested in getting those into the list, could you open one pull request per service?

This is not merge-able in its current form (references removed categories like "Web Hosting", PNGs are actually 404s, too many services), unfortunately, so closing.

PR #1819, opened 2017-10-10
Quick update of zh-TW translation

PeterDaveHello:


PeterDaveHello:
@vyp @nylira @strugee @Zegnat please take a look please, thank you!

vyp:
Thank you for taking the time to do this! :sparkles:

PeterDaveHello:
I'll send another PR to try to complete the translation when I have more time soon, thanks.

PeterDaveHello:
@vyp may I ask how long should I expect to have it deployed? Thanks.

vyp:
Depends on nylira and when they can manually build and deploy, because he's the only one with access to the actual site. The `public` directory at https://github.com/nylira/prism-break-static should reflect the current state of the site. Usually the time it takes can be anywhere from a day to about a week or so, once it goes into `master`.

And yes that would be great of course, thank you in advance! But even partial translations are better than nothing, so if you end up not finding the time to complete it all, sending whatever you have done would still be appreciated!

PeterDaveHello:
That's sweat :)

nylira:
A new version will be deployed shortly. Thanks guys!

strugee:
@nylira somehow I thought you built/deployed on a daily cronjob? Is that not the case?

vyp:
@strugee it is not as far as I know. Due to the long build time and in the past the frequent number of commits, it was chosen to simply handle it manually.

PR #1820, opened 2017-10-12
Fix Autistici hyperlink

Evanito:
The current link is a 404, my change fixes that.

### DIFF
--- a/source/db/en-projects.json
+++ b/source/db/en-projects.json
@@ -151,7 +151,7 @@
     "source_url": "",
     "name": "Autistici/Inventati",
     "tos_url": "",
-    "url": "http://www.autistici.org/",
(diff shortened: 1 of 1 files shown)


vyp:
Thank you. While we're at it, could you make it https://www.autistici.org/ instead?

Evanito:
The `/en/` didn't change anything functionally and I opted to keep it because it was specifically listed under the en list.

vyp:
Yeah I saw, I would have opted to keep `/en/` if the other locales were also linkable statically, instead of being managed by javascript, meaning that all the other languages just point to `#` and you have to click for them after reaching the site. Also I want it to be a https link if you could. :+1:

Evanito:
Didn't notice that the first time.

Issue #1055, opened 2014-07-05
Add Firefox Sync (own server) to File Storage & Sync


andr3jx:
I want to suggest to add FIrefox Sync to File Storage & Sync. You can use it to sync bookmarks, passwords, history & more. 
The data is encrypted in the client and the server can't decrypt it. We already have Firefox in the list but we should add that this Sync service is secure and that it is possible to setup your [own sync server](https://github.com/mozilla/fxa-auth-server).

PR #1823, opened 2017-10-20
Add xBrowserSync

nero120:
xBrowserSync is a free and open-source tool for syncing bookmarks (initially, additional browser data syncing will be added in future) across browsers and mobile devices. It would be great to have it added to prism break!


strugee:
Hi @nero120, is there a compelling advantage xBrowserSync provides over the tools already listed? Remember, PRISM Break is about quality, not quantity.

nero120:
Hi @strugee, the reason I started xBrowserSync was that I wanted to completely stop using big tech provider services, but that I could not find a compelling alternative to Google Chrome's bookmark/browser data sync. I designed xBrowserSync to be anonymous and usable "out of the box", that is, there is no submission of personal data required to start using the service, and once you've downloaded the extension/app it's a one click sync. The synced data is encrypted (using AES) client-side and the user's encryption key (password) is never transmitted, so the data is secure. xBrowserSync is also FOSS, and I don't intend to ever charge for the service. xBrowserSync is currently available for Chrome (and uses the bookmarks API so bookmarks are native and the browsers bookmarking features are fully supported), and apps are available for Android & iOS. I'm currently working on Firefox support.

None of the listed PRISM Break services provide all of these features and that's really why I started the project. xBrowserSync has a growing user base and it's actively helping people to de-microsoft or de-google and take back control of their own data. I believe it would make a valuable addition to the PRISM Break directory. :)

yegortimoshenko:
Currently, Firefox is not supported, only Chrome: https://www.xbrowsersync.org/

Additionally, the last commit was half a year ago: https://github.com/xBrowserSync/App/commits/master

I suggest to close this for now and re-evaluate later if anything changes.

nero120:
I am the repo owner and the project is actively maintained, i am about to start work on the firefox extension which has been requested relatedly. I had a break for the last six months whilst i got married and focused on other commitments after spending a year in the development of this project.

I ask you not to close this request, i am currently the only dev and you understand that people who have jobs and lives cannot spend all their time working on FOSS projects. xBrowserSync is very much alive and the user base grows by the day.

-------- Original Message --------

strugee:
That seems reasonable. Congrats on your marriage @nero120!

nero120:
Thank you @strugee!

yegortimoshenko:
I still think it should not be listed, even if it were maintained.

We only list Firefox-based browsers, that already come with end-to-end encrypted sync for history, passwords, browser settings, etc. My opinion is that this can only be properly done by privileged code and implementing such functionality via an add-on is inferior, both in integration with browser and very limited privileges. 

Also, given that I can't find any pinning code on either [client](https://github.com/xBrowserSync/App) or [server](https://github.com/xBrowserSync/API) it seems that this completely relies on HTTPS for security, with all of its problems (PKI, downgrade attacks). It should adopt trust on first use (like ssh) to be comparable to Firefox Sync in terms of quality.

Also, client heavily couples UI code and logic (model and view): https://github.com/xBrowserSync/App/blob/master/js/app.js

Also see #1055 and https://blog.mozilla.org/services/2014/02/07/a-better-firefox-sync/

nero120:
@yegortimoshenko you are of course entitled to your opinion, but please try not to spread disinformation. I'm happy to receive criticism but at least try and get your facts straight.

The point of xBrowserSync is to help people *avoid* corporations that are monolithic, opaque and anti-privacy. In my opinion, running straight into the arms of the Mozilla corporation is not exactly my idea of  helping people to "opt out of global data surveillance programs like PRISM, XKeyscore and Tempora", or "help make mass surveillance of entire populations uneconomical! We all have a right to privacy, which you can exercise today by encrypting your communications and ending your reliance on proprietary services.". This is the same Mozilla corporation that has [committed itself](http://uk.businessinsider.com/mozilla-new-initiative-counter-fake-news-2017-8?r=US&IR=T) to combating what it declares is "fake news" (or simply "censorship" to normal people), using software developed by Full Fact foundation, which in turn is backed by eBay founder Pierre Omidyar and George Soros. Not the kind of people I would like to trust with my data.

Also, xBrowserSync does not "completely rely on HTTPS", being that all data is encrypted client-side using AES (via [crypto-js](https://github.com/brix/crypto-js)). Even still, if your criticism is that HTTPS is not a secure enough protocol for your standards then I trust you don't use the web at all??

Also, see [here](https://palant.de/2018/03/13/can-chrome-sync-or-firefox-sync-be-trusted-with-sensitive-data) for an up to date criticism of Firefox Sync security:

```
There is an important weakness here nevertheless: scrypt runs on the Firefox Accounts server, not on the client side. On the client side this protocol uses PBKDF2-HMAC-SHA256 with merely 1000 iterations. And while the resulting password hash isn’t stored on the server, if somebody can read it out when it is being transmitted to the server they will be able to guess the corresponding password comparably easily. Here, a single Nvidia GTX 1080 graphics card could validate 1.2 million guesses per second. While the effort would have to be repeated for each user account, testing 1.5 billion known passwords would be done within twenty minutes. And a 40 bits strong password would fall within five days on average. Depending on what’s in the account, spending this time (or adding more hardware) might be worth it.
```

On comparison, how long do you think it would take you to break the AES encryption used to encrypt xBrowserSync data?

Finally, you are incorrect in stating "client heavily couples UI code and logic". If you understood the code you'd see that app.js is purely UI. The app logic is contained in other files. Besides, I've built and maintained hundreds of applications in my career, the lack of separation of presentation and logic barely even registered as a practical issue. Developers use whatever approach works for them, patterns are not orthodoxy and we are not cult members.

yegortimoshenko:
> On comparison, how long do you think it would take you to break the AES encryption used to encrypt xBrowserSync data?

https://github.com/xBrowserSync/App/blob/d8d788ea2596f8fc8e15ab6328d12aca6ae32521/js/utility.js#L27

Oh, it looks like it uses ECB mode :-) There is nothing to break; unfortunately, your data is not encrypted. Any adversary who gets access to your server (or yourself) can tell which sites users have visted by statistical analysis. Here is an explanation with a good visualization: https://crypto.stackexchange.com/questions/20941/why-shouldnt-i-use-ecb-encryption

Just AES is equivalent to AES ECB, ECB is essentially block passthrough where other modes embed IVs, authentication, etc.

Besides that, anyone with access to the server can purposefully corrupt some blocks to meaningfully change data in others: always use authenticated encryption (for AES, modes like GCM, OCB) unless you have clear reasons not to.

nero120:
Crypto-js [by default uses](https://github.com/brix/crypto-js/blob/develop/docs/QuickStartGuide.wiki#Block_Modes_and_Padding) CBC block mode and Pkcs7 for padding, which is what xBrowserSync in turn uses.

It's actually on my to-do list to switch from using crypto-js to using the Web Cryptography API as it's commonly supported now.

yegortimoshenko:
Perhaps I am wrong. But if this is CBC, where's the IV handling? I don't see any, there is just a missing argument in API that you consume. Without IV, CBC is no better than ECB.

yegortimoshenko:
Also, you've quoted a passage about key stretching used by Firefox Sync (1000 rounds of PBKDF2). I don't see any key stretching in your code, that would mean bruteforcing AES with non-stretched key would be about an order of magnitude faster than in case of Firefox, and you seem to directly use password as AES key instead of a proper 128-bit AES key. Users' passwords are 40-50 bits on average, so it would probably be trivial to bruteforce most users' keys very quickly in that case.

yegortimoshenko:
And CBC is still not authenticated, so danger of purposeful profile corruption (e.g. to replace bookmarks with malicious counterparts) still stands.

nero120:
> Perhaps I am wrong. But if this is CBC, where's the IV handling? I don't see any, there is just a missing argument in API that you consume. Without IV, CBC is no better than ECB.

You'd have to dig in to the crypto-js code.

> Also, you've quoted a passage about key stretching used by Firefox Sync (1000 rounds of PBKDF2). I don't see any key stretching in your code, and bruteforcing AES with non-stretched key would be at least an order of magnitude faster than in case of Firefox. Given that xBrowserSync seems to directly use password as AES key instead of a proper 128-bit AES key, and users' passwords are 40-50 bits on average, it would probably be feasible to bruteforce most users' keys in minutes.

Again, not true. The plain text passphrase that xBrowserSync passes into the `CryptoJS.AES.encrypt` function is used by crypto-js as a salt to generate a 256-bit key which is in turn used as the AES encryption key. I would have thought that breaking the resulting encrypted string would be much harder than brute forcing the weaker password hash generated by Firefox Sync client side.

My point is not to imply that Firefox Sync is not suitable as a syncing tool, merely that it is not a reason not to admit xBrowserSync to prism break. Again, xBrowserSync was created for users who value their privacy and do not want to trust their data (even encrypted) to corporations.

yegortimoshenko:
> Again, not true. The plain text passphrase that xBrowserSync passes into the CryptoJS.AES.encrypt function is used by crypto-js as a salt to generate a 256-bit key which is in turn used as the AES encryption key.

Transformation has to stay constant, i.e. same password leads to the same key. If key derivation function itself in this case is not computationally expensive (like scrypt), there is little hope, because attacker won't have to go through all 256-bit keyspace, they can constrain themselves to tiny 40-50 bit space that people use for passwords.

CryptoJS uses EVPKDF, deprecated OpenSSL key derivation function, that uses MD5 for hashing and performs just one round: https://github.com/CryptoStore/crypto-js/blob/3.1.2/src/evpkdf.js#L19

Here is EVPKDF used in key derivation mechanism for AES that you've described: https://github.com/CryptoStore/crypto-js/blob/3.1.2/src/cipher-core.js#L760

Then compare it to Firefox Sync. They use it for the same purpose, to derive key from a password, but use PBKDF2 with 1000 rounds, and SHA256 as a hashing function.

yegortimoshenko:
And to be clear, Firefox Sync can certainly be critiqued for not using bcrypt or scrypt for key derivation, but using one round of some old key derivation algorithm using MD5 is significantly worse than 1000 rounds of PBKDF2+SHA-256.

yegortimoshenko:
Situation is no different for repo that you're using:

https://github.com/brix/crypto-js/blob/release-3.1.2/src/cipher-core.js#L760
https://github.com/brix/crypto-js/blob/release-3.1.2/src/evpkdf.js#L22

The problem is that the passage that you've quoted about Firefox Sync using weak key derivation algorithm applies many times over to xBrowserSync. Firefox Sync, just like xBrowserSync, derives AES key from the password, but does so via 1000 rounds of PBKDF2: https://github.com/mozilla/fxa-auth-server/wiki/onepw-protocol

Aattacking this PBKDF2 requires active eavesdropping. If one only has access to server's DB, they would have to attack scrypt, which is pretty much impossible.

That said, there are other problems to xBrowserSync: unauthenticated encryption allows server to maliciously corrupt profile, and it fully relies on HTTPS to determine server identity while Firefox Sync does cert pinning.

nero120:
@yegortimoshenko the point is you're criticising xBrowserSync (which is fine) whilst advocating another sync that exhibits a similar weakness but more worryingly is controlled by an organisation that is anti-privacy and in direct conflict with the stated objective of PRISM BREAK. I do not ask for or collect *any* personal data from users, Mozilla ask you to open an account, collect your personal info and associate all of your data with it under one insecure master password. In any case, I (or anyone who wants to submit a pull request) can resolve the issues with the encryption xBrowserSync uses relatively easily, how long will it take Mozilla to address this issue, and do they even want to fix it?

Fundamentally this is about trust. I do not expect users of xBrowserSync to trust me, that is why the code is open source and they can run their own service to sync the client to if they want to remove all doubt. Firefox Sync doesn't even begin to offer that level of autonomy, if it did I never would have built xBrowserSync.

yegortimoshenko:
Firefox Sync can be self-hosted, and it's free software too.

nero120:
Ok fair point. Do you self host it?

yegortimoshenko:
I'm not sure if that changes anything, but I don't use any sync service currently. 

I'm not sure what in particular in this discussion you want to point out, but if one wants to host their own Firefox Accounts service, here's documentation: https://moz-services-docs.readthedocs.io/en/latest/howtos/run-fxa.html

Zegnat:
To drag this issue back on topic: we should decide if it is worth recommending xBrowserSync and what proprietary products (Xmarks, perhaps?) we are recommending it over. That’s the only thing PRISM Break is concerned with.

If @yegortimoshenko has detected cryptographic weaknesses in xBrowserSync, those should probably be filed on the respective repository. Those can then be linked to from this issue as reasons to exclude inclusion. If true (I haven’t checked) that xBrowserSync doesn’t do any key stretching on the password it uses for the AES encryption that is a bad practice that needs to be addressed, no matter what Firefox Sync does in comparison…

If it is decided that a browser sync category is interesting and needed at all, it is easy enough to start a discussion on whether Firefox Sync should go under proprietary, or be recommended alongside xBrowserSync as a free option. Not sure any other discussion of Firefox Sync on this specific issue is of any consequence.

yegortimoshenko:
> If @yegortimoshenko has detected cryptographic weaknesses in xBrowserSync, those should probably be filed on the respective repository. Those can then be linked to from this issue as reasons to exclude inclusion.

I don't see why: perhaps they should be, but that's of no concern to PRISM Break. Cryptographic weakness is grounds for exclusion no matter whether it has been filed against project's issue tracker.

> Not sure any other discussion of Firefox Sync on this specific issue is of any consequence.

It supposedly is, Firefox Sync doing stronger crypto and also self-hostable means there is better software that does the same job, plus much more (quite a few things that xBrowserSync will never be able to do, because of not being privileged).

We try to list quality software, and quality only means something in comparison to other products. That's why we've decided to drop YaCy, for example: yes, it is free, but search quality is so bad in comparison to other (proprietary) options that it's not worth listing.

yegortimoshenko:
I'd like to veto against this based on cryptographic issues noted above (CBC, key derivation function that does one round of MD5).

nero120:
This is in the process of being resolved, the app will also be updated to replace the use of the cryptojs library with [Window.crypto](https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto).

nero120:
Just wanted to update you guys, I've recently committed an update for [v1.4.0](https://github.com/xBrowserSync/App/tree/v1.4.0) (upcoming major update that includes support for Firefox) that resolves the issues @yegortimoshenko highlighted in this thread:

- Replaced crypto-js with the native [browser implementation](https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto) of the [Web Crypto API standard](https://www.w3.org/TR/WebCryptoAPI/).
- Uses AES-GCM for the encryption algorithm.
- Uses PBKDF2 with 100,000 rounds of SHA-256, a random 16 byte IV and the user's random 32 char sync ID as a salt.

Users are automatically upgraded to using the above encryption method when they retrieve their existing sync with v1.4.0. You can check out the relevant code [here](https://github.com/xBrowserSync/App/blob/v1.4.0/js/utility.js), the methods are `encryptData` and `decryptData_v2`.

I'm currently fixing up the remaining Firefox bugs and will hopefully release v1.4.0 within the next few weeks.

Also, since Xmarks [shut down](https://www.ghacks.net/2018/03/31/logmein-to-shut-down-xmarks-on-may-1-2018/) I've been inundated with user requests to support Firefox. In light of the lack of independent, open source browser syncing tools, xBrowserSync is a very relevant addition to PRISM Break and I hope you decide to include it once v1.4.0 is released.

alerque:
Thanks for the update and addressing those issues. Please ping back here when it's released and we'll give it consideration. For the moment this is still tagged as blocked. I'll be reviewing it for my own personal use as well as PRISM-Break when its ready for FF.

nero120:
No worries, will do @alerque.

PR #1840, opened 2017-12-02
Add specific security inclusion guidelines

### LINKED ISSUE #1685: README.md doesn't properly communicate inclusion criteria
See https://github.com/nylira/prism-break/pull/1683#issuecomment-279193318 - the README.md file's project inclusion guidelines doesn't really do a great job at communicating e.g. our policies on project security. We should make those things explicit.

Here's the rough list I came up with for our security guidelines:

* Understand security and take it seriously (this is very broad, but can range from "don't invent your own crypto" to "think about and design against metadata leaks")
* Offers security advantages that other projects don't (this is why e.g. Sandstorm's containerization model is particularly interesting to us)
* Can keep up with security events
* Doesn't depend on, imply or promote insecure things including and especially things in the "proprietary" column

strugee:
Fixes #1685 

Would love some feedback on this - I feel like this list isn't really good enough but I'm not sure where to go from here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,9 @@ PRISM Break strives to promote the best open source applications. Ease of use, s
 
 This can vary case-to-case, but generally here are the attributes we look for:
 
+ ** Can keep up with security events
  ** Understand security and take it seriously (this is very broad, but can range from "don't invent your own crypto" to "think about and design against metadata leaks")
  ** Offers security advantages that other projects don't
- ** Can keep up with security events
- ** Doesn't depend on, imply or promote insecure things including and especially things in the "proprietary" column
 
  - **Before suggesting software, please first search this repository to see if your request has already been made.**
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ PRISM Break follows [the GNU/FSF definition of Free Software](https://www.gnu.or
 
 PRISM Break strives to promote the best open source applications. Ease of use, stability, and performance matter. This is the first time many people are looking to leave their proprietary walled gardens. Let's make it a good experience for them. If you're writing a privacy-minded FOSS app, please finish it before asking PRISM Break to promote it.
 
- - **The software must actually be secure.**
+ - **The software must have a strong security focus.**
 
 This can vary case-to-case, but generally here are the attributes we look for:
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,15 @@ PRISM Break follows [the GNU/FSF definition of Free Software](https://www.gnu.or
 
 PRISM Break strives to promote the best open source applications. Ease of use, stability, and performance matter. This is the first time many people are looking to leave their proprietary walled gardens. Let's make it a good experience for them. If you're writing a privacy-minded FOSS app, please finish it before asking PRISM Break to promote it.
 
+ - **The software must actually be secure.**
+
+This can vary case-to-case, but generally here are the attributes we look for:
+
+ ** Understand security and take it seriously (this is very broad, but can range from "don't invent your own crypto" to "think about and design against metadata leaks")
+ ** Offers security advantages that other projects don't
+ ** Can keep up with security events
+ ** Doesn't depend on, imply or promote insecure things including and especially things in the "proprietary" column
+
  - **Before suggesting software, please first search this repository to see if your request has already been made.**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,15 @@ PRISM Break follows [the GNU/FSF definition of Free Software](https://www.gnu.or
...
 PRISM Break strives to promote the best open source applications. Ease of use, stability, and performance matter. This is the first time many people are looking to leave their proprietary walled gardens. Let's make it a good experience for them. If you're writing a privacy-minded FOSS app, please finish it before asking PRISM Break to promote it.
 
+ - **The software must actually be secure.**
+
+This can vary case-to-case, but generally here are the attributes we look for:
+
+ ** Understand security and take it seriously (this is very broad, but can range from "don't invent your own crypto" to "think about and design against metadata leaks")
+ ** Offers security advantages that other projects don't
+ ** Can keep up with security events
+ ** Doesn't depend on, imply or promote insecure things including and especially things in the "proprietary" column
```
hasufell: Don't confuse insecure with proprietary. I'd guess a lot of proprietary software is actually more secure than some opensource counterparts (if there are any, often there is not for industry grade tools). We just don't want it because of **trust**. I'm not sure we really need to say again that we don't like proprietary software. That's already established.

README.md
```diff
@@ -18,6 +18,15 @@ PRISM Break follows [the GNU/FSF definition of Free Software](https://www.gnu.or
 
 PRISM Break strives to promote the best open source applications. Ease of use, stability, and performance matter. This is the first time many people are looking to leave their proprietary walled gardens. Let's make it a good experience for them. If you're writing a privacy-minded FOSS app, please finish it before asking PRISM Break to promote it.
 
+ - **The software must actually be secure.**
+
+This can vary case-to-case, but generally here are the attributes we look for:
+
+ ** Understand security and take it seriously (this is very broad, but can range from "don't invent your own crypto" to "think about and design against metadata leaks")
+ ** Offers security advantages that other projects don't
+ ** Can keep up with security events
```
hasufell: This is the best point and also relatively easy to evaluate. It should be at the top of the list.

hasufell:
I'd add something like

* applies special security methodologies (e.g. formal, semi-formal or workflow methods that are security relevant)

README.md
strugee: Done

README.md
strugee: Right, that's what I meant but I was conflating the language. In any case I think you're right that this is just cluttering up the list; will remove

strugee:
> applies special security methodologies (e.g. formal, semi-formal or workflow methods that are security relevant)

It seems to me like this is included in "Offers security advantages that other projects don't"? That was the intention at least, IIRC

strugee:
@hasufell ping - what do you think of these changes?

README.md
```diff
@@ -18,6 +18,14 @@ PRISM Break follows [the GNU/FSF definition of Free Software](https://www.gnu.or
 
 PRISM Break strives to promote the best open source applications. Ease of use, stability, and performance matter. This is the first time many people are looking to leave their proprietary walled gardens. Let's make it a good experience for them. If you're writing a privacy-minded FOSS app, please finish it before asking PRISM Break to promote it.
 
+ - **The software must actually be secure.**
```
hasufell: I'd loose the description here, because it's hard to evaluate whether something is actually secure. It's better to say something along these lines: **Must have a strong security focus**

sim-s-singh:
Could we not also give some label or symbol to show software that has had a 3rd party code review? A strong audited security base is critical for privacy/security/anonimity centric applications.

strugee:
> Could we not also give some label or symbol to show software that has had a 3rd party code review?

@sim-s-singh I believe we track this information in JSON files but I'm not sure if we actually expose that information in the website.

That being said that's a separate issue. Can you please file a new issue about it, after searching for duplicates of course? Thanks!

strugee:
@hasufell pushed the update you suggested.

Looking at it now though I wonder if we'd be better off saying "must be reasonably secure"? The point I'm trying to get across is that we look directly at the actual evidence of the project's implementation and practices, not just what they focus on or whatever. Perhaps "reasonably secure" gets that across while sidestepping the issue of us not being able to 100% declare something "secure".

PR #1841, opened 2017-12-02
Remove identi.ca note

### LINKED ISSUE #1839: Remove identi.ca note
identi.ca is mentioned in https://prism-break.org/en/all/#social-networks but registration is closed since it effectively turned into a super-node. Even if registration is open it we should still remove it because super-nodes are an antipattern. So we should remove the note about it.

mimi89999:
Closes #1839


strugee:
This seems to have gotten rid of more than the identi.ca note.

alerque:
I think this PR can be closed as obsolete. I must have missed that it was even here but at some point I (re?)did this task in #1855 and that's been merged. My apologies @mimi89999 for stepping on your work starting this. Normally I would have helped fix up previous contributions rather than redo them. Your PR was even linked from the issue so I'm not sure how I missed it.

strugee:
Ah, I'd forgotten about this too. Thanks @mimi89999, and sorry!

PR #1852, opened 2018-01-04
Remove obsolete Random Agent Spoofer

### LINKED ISSUE #1815: Get rid of extensions that won't support Firefox 57+
Self-Destructing Cookies is already mentioned [here](https://github.com/nylira/prism-break/issues/1796).

RAS is also [no longer maintained](https://github.com/dillbyrne/random-agent-spoofer#readme) and will not work under Firefox 57 that will come out on November 14.

And while we're at it, the Disconnect extension is also no longer relevant because their malware and spyware filters are included in Firefox and uBlock Origin (under the 3rd party filters).

alerque:
Closes #1815, see issue for discussion.
 
~~Note this PR one whitespace cleanup commit not related to this function — but that makes these sort of removals much easier. This can/will conflict if other PRs are merged first. Doing these removals manually is painful compared to automating it with a JSON filter. I can easily redo this if other PR merges break it before it get merged, just ping me here.~~

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -4348,47 +4348,6 @@
   },
   {
     "development_stage": "released",
-    "description": "A privacy enhancing firefox add-on which aims to hinder browser fingerprinting.",
-    "license_url": "https://github.com/dillbyrne/random-agent-spoofer/blob/master/LICENSE",
-    "logo": "ras.png",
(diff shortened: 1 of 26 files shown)


alerque:
Rebased now that the big whitespace cleanup commit is in master.

PR #1853, opened 2018-01-04
Remove xabber

### LINKED ISSUE #1659: Remove Xabber?
The last commit to [xabber-android](https://github.com/redsolution/xabber-android) was on 28th December 2015 (over a year ago). Should it still be recommended?

alerque:
Closes #1659, see extensive discussion there.

~~Note this PR has a duplicate commit with #1852. Whichever one is merged first will suffice, the second one can be rebased. Doing these removals manually is painful compared to automating it with a JSON filter. I can easily redo these if other PR merges break them before they get merged, just ping me here.~~

### DIFF
--- a/source/db/en-projects.json
+++ b/source/db/en-projects.json
@@ -686,7 +686,7 @@
     "description": "Conversations is an open source XMPP/Jabber client for Android 4.0+ smart phones.",
     "license_url": "https://github.com/siacs/Conversations/blob/development/LICENSE",
     "logo": "conversations.png",
-    "notes": "",
+    "notes": "Conversations' one native code dependency on iNPUTmice is [discussed here](https://github.com/nylira/prism-break/issues/1659#issuecomment-269921269).",
     "privacy_url": "",
     "source_url": "https://github.com/siacs/Conversations.git",
     "name": "Conversations",
--- a/source/db/en-projects.json
+++ b/source/db/en-projects.json
@@ -6124,33 +6124,6 @@
   },
   {
     "development_stage": "released",
-    "description": "OTR-encrypted instant messaging for Android.",
-    "license_url": "https://github.com/redsolution/xabber-android/blob/master/LICENSE",
-    "logo": "xabber.png",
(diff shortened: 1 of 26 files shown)


source/db/en-projects.json
```diff
@@ -714,7 +714,7 @@
     "description": "Conversations is an open source XMPP/Jabber client for Android 4.0+ smart phones.",
     "license_url": "https://github.com/siacs/Conversations/blob/development/LICENSE",
     "logo": "conversations.png",
-    "notes": "",
+    "notes": "Conversations' one native code dependency on iNPUTmice is [discussed here](https://github.com/nylira/prism-break/issues/1659#issuecomment-269921269).",
```
strugee: Do other localizations have a `notes` key for Conversations? If so they need to be removed.

strugee:
Nevermind, looks like it was a transient network problem. I kicked Travis, should show up green soon.

source/db/en-projects.json
alerque: @strugee If I understand this the commit 226d8a4  on my #1854 PR will take care of removing all blank note fields across all listings. If that's not correct then I'm not quite sure I understand what's going on.

alerque:
Rebased now that whitespace cleanup and empty notes removal commits are in master.

source/db/en-projects.json
strugee: @alerque I was wondering if there's a `notes` field for other Conversations listing that somehow got out of sync and had text in it (and therefore wasn't picked up by your commit). I would assume not but it'd be nice to check. Seems pretty unlikely though so if you don't have time to check I'd be fine merging this as-is :thumbsup:

alerque: @strugee I just ran a query on that field and there are no other `.notes` keys with obsolete content on Conversations listings in any language, so no ripping out just the blanks didn't miss any in this case. Good thought though.

strugee: Great, thanks for checking! LGTM!

PR #1854, opened 2018-01-04
Drop note about FISH

alerque:
See note on #1852 about whitespace commit. I can rebase this easily if/when it becomes a merge conflict, just ping me.


alerque:
Okay, I'll update with that in mind.

strugee:
@alerque thanks so much :tada:

alerque:
@strugee I rebased this and then dropped all the empty note fields in project files except for the English one. I figured if this logic was sound about pass-through to EN if no translation present it must work for all listings right? If I'm missing something there let me know.

strugee:
> I figured if this logic was sound about pass-through to EN if no translation present it must work for all listings right?

Yeah, if there's _any_ key missing for a given project, it will fall back to the English data for that particular key. Dropping all the empty ones sounds like a great plan.

alerque:
@strugee In that case shouldn't we drop _all_ the empty keys in translation files, not just `notes` fields? I actually did this originally and then second guessed myself and backed off to just notes for this PR. If we want all of them gone I'll submit that.

strugee:
Ah yeah, that sounds great.

Issue #1633, opened 2016-12-13
Building takes very long (slow contribution workflow)


hasufell:
Contributing has become slightly annoying, because of 2 things:

1. checking/manipulating all language files
2. building all of them takes a long time, even on a very powerful machine

I've only done haskell web development, so my suggestion would be [hakyll](https://jaspervdj.be/hakyll/), which can use similar human-readable formats for the actual text.

strugee:
1. will be solved by #1627. As for 2... not sure. I might bust out a profiler if I have time and try to speed things up.

strugee:
#1639 landed which allows parallel jobs. Do we still need to keep this open?

alerque:
The build still takes a long time and some tooling for simple repetitive tasks across across languages would be nice, but the build is a lot faster than it used to be (and includes some checks) and baring a rewrite from scratch I don't see it changing dramatically soon. I've been playing around with using `jq` to automate common tasks (see my recent commit messages) but don't have anything setup to share yet.

In any case I'm not sure this issue is serving a purpose any more. If somebody wants to propose a rewrite that's fine, and more specific issues for any commands that would help contributors would be welcome, but this issue is kind of mushy with no real action to take.

strugee:
Yeah, I agree. I wonder if it would help to put a note in either the Pull Request template or `CONTRIBUTING.md` or something like that to remind people that if they run into something specific that's difficult while contributing, they should tell us?

hasufell:
> to remind people that if they run into something specific that's difficult while contributing, they should tell us?

I just did. So why should this issue be closed?

strugee:
> I just did.

Where, in this issue? I don't think there's anything that's specific enough to actually be actionable.

hasufell:
> Where, in this issue? I don't think there's anything that's specific enough to actually be actionable.

First, long build time is **very** specific. Second... just because you don't know how to fix an issue, doesn't make it invalid.

alerque:
> First, long build time is very specific.

"Long" and "very long" are rather subjective terms.  But if I check out this project from the date you opened the issue a full build (not included downloading NPM modules and such) clocks in at 17:02. The current HEAD clocks in at 5:21 on the same machine.

[...]

PR #1867, opened 2018-01-31
Fix build process for safer deploys

### LINKED ISSUE #1863: 404 Not Found on PRISM Break homepage
https://prism-break.org

alerque:
* Use project local binaries in preference to system ones, closes #1863.
* Cut another hunk of time out of my contributor workflow, see #1633. Initial lint time is down from 60 seconds to 12, subsequent runs only check changed files.
* Fix broken lint target (and bogus JSON it should have caught before!).
* Move a step closer to deterministic builds using yarn.


alerque:
@strugee Was there anything else I said I was going to do with the build process that doesn't  have an open issue? I realize I was going to strip empty and untranslated values from language files, but I think that's a job for another PR. Anything else I'm forgetting about that has cropped up recently?

strugee:
Whoops, sorry I missed this @alerque. I don't think there was anything else but I honestly don't remember at this point. Thanks for working on this.

As I've said before [I don't like yarn](https://twitter.com/strugee2/status/882807134110457856), especially because they [don't play nice with others](https://twitter.com/maybekatz/status/954113152496513024). I can definitely live though, especially since this is already merged. :thumbsup:

alerque:
@strugee I'm well aware of the drama surrounding these two projects. I also hear you loud and clear on the principle of open source. "Not made here" syndrome is a real thing and certainly applies to Yarn. But at the same time I don't think it's fair to hold up NPM as some bastion of good open source. In fact sometimes the project [can be a downright dumpster fire](https://github.com/npm/npm/issues/19883) and when security is on the line I think Yarn is actually a better managed project. The fact that they didn't just fork and contribute upstream is concerning too, but it's not like NPM has a track record of being friendly is this regard. Go look through the accepted vs. rejected PR's for each project, then at the issue trackers for each.

From a pragmatic standpoint when usability and security are on the line at this point I'd rather use `yarn`. Of course I wish it wasn't piloted by a corporate sponsor and the developer community hadn't split so badly, but I can also understand why they felt they couldn't be held up trying to corral the cats and just forged ahead with something. And the competition has helped, NPM has gotten it's act together in a lot of ways that it didn't when there wasn't that pressure. Even the issue of proper locking and reproducible builds only got better on the NPM side after people started jumping ship like crazy when it became obvious that Yarn did that better. It had been talked about as a problem forever but nothing was getting done.

At the end of the day I don't want to get into this much more and will probably drop out of even discussing it here. If there in a convincing pragmatic reason going the other way and using `npm` would make the build process here better/safer then I'm certainly willing to hear it. My main beef was that it wasn't safe to keep supporting both in parallel and foregoing any reproducibility. Since I had in mind a safer, faster, and more reproducible build system I picked the tool I trusted more for those aspects.

Issue #1723, opened 2017-04-02
Mastodon


hamoid:
Hi, any thoughts on https://mastodon.social/about ? I just heard about it. Cheers!

Atavic:
:+1: for the source on GH.
:-1: for the Terms (everytime, terms provided are about the webpage and forum, not the service)

Unless I didn't spot them, I couldn't find the User _Terms for the service_ itself.

strugee:
Terms of Service are mostly irrelevant because we'd be recommending the software, not the primary hosted instance.

Hillside502:
Open Source Mastodon Alternatives - AlternativeTo.net
https://alternativeto.net/software/mastodon/?license=opensource

strugee:
Mastodon ~= GNU Social, which is already listed. The main questions here are Mastodon's licensing (though I can't imagine it's problematic) and development community which has an impact on how security will be treated.

Gargron:
Mastodon is AGPLv3 licensed

- https://github.com/tootsuite/mastodon
- https://joinmastodon.org

>Mastodon ~= GNU Social, which is already listed

This is not quite true. Even though they share the same network, it's wildly different software with different design and features. Now it's also primarily ActivityPub-based, which is a new federation protocol.

strugee:
> This is not quite true. Even though they share the same network, it's wildly different software with different design and features. Now it's also primarily ActivityPub-based, which is a new federation protocol.

Lol, I'm aware - I'm `ajordan` on W3C IRC :P

FWIW when I wrote that comment, it was much more accurate.

Thanks for the ping. Since the Social Networking section is kinda overloaded, I vote for replacing GNU Social with Mastodon. AFAIK Mastodon is easier to set up and ActivityPub is a W3C standard that lots of these projects are converging on, so Mastodon will interoperate with more stuff. Plus it supports things like PMs that GNU Social does not and development seems more active as well. @Gargron does it still federate with GNU Social instances or did y'all drop that already? That would be another compelling point in favor of Mastodon.

deutrino:
It still federates with ostatus-capable nodes, including gnusocial.

strugee:
Nice. I don't see any reason to prefer GNU Social over Mastodon then, and I see compelling reasons to prefer Mastodon. Marking as accepting PRs.

alerque:
[...]

PR #1873, opened 2018-02-15
Add Mastodon

outloudvi:
Also see #1723.
Pictures are modified from the official [Press Kit](https://joinmastodon.org/press-kit.zip). I'm not really good at doing svgs so feel free to put a better version there.


yegortimoshenko:
I think GNU Social should be removed in favor of Mastodon. Both are implementations of the same protocol suite, and Mastodon is clearly more vibrant and much more active.

Compare: 

* https://git.gnu.io/gnu/gnu-social/commits/master
* https://github.com/tootsuite/mastodon/commits/master

alerque:
@yegortimoshenko Lets please keep the discussion for this on the related issue (#1723) rather than splitting it across locations. Discussion on PR should generally be about the actual implementation of the changes rather than whether or not to make a change which we usually agree on ahead of time in issues. If there is no issue first then of course discussing in a PR is fine, but not in this case. Thanks.

yegortimoshenko:
This, of course, directly applies to the implementation of the changes. This pull request doesn't come with a GNU social removal.

> That being said I don't think you said anything that hasn't been discussed and wasn't already our conclusion.

There has been no clear consensus, or conclusion, on what to do with GNU social.

alerque:
...which is why that should be discussed on the issue until there IS a consensus, _then_ brought here for implementation. As it stands I think the PR does reflect what consensus we do have to date.

yegortimoshenko:
#1873 might be applicable here, depending on its resolution. But that of course can be fixed later.

outloudvi:
@yegortimoshenko Agree. Mastodon seems to be at the state of what Up1 was when there is a official demo Up1 site. Or it might be delayed because these "demo" sites are active, with lots of users in and thus unlikely to close suddenly.

yegortimoshenko:
I think mastodon.social is a "web service", but Mastodon isn't, and up1.ca is a "web service", but Up1 isn't. While it's not possible to suggest one without suggesting the other, the same applies to Nextcloud, OpenVPN, even Tor.

It's also important to try to have as much federation as possible, or we would end up with Gmail situation, where most email goes through Gmail because sender or at least one of recipients probably uses Gmail. I think "web service" tag promotes just hopping on someone else's instance, but perhaps I'm over-analyzing that.

source/db/en-projects.json
```diff
@@ -6254,5 +6254,37 @@
       }
     ],
     "slug": "android-password-store"
+  },
+  {
+    "development_stage": "released",
+    "description": "Self-hosted, decentralized OStatus-compatible social network.",
+    "license_url": "https://github.com/tootsuite/mastodon/blob/master/LICENSE",
+    "logo": "mastodon.svg",
+    "notes": "Policies may vary between instances, choose one that fits you.",
```
strugee: Can we add a note here encouraging people to self-host?

source/db/en-projects.json
```diff
@@ -6254,5 +6254,37 @@
...
+    "privacy_url": "",
+    "source_url": "https://github.com/tootsuite/mastodon",
+    "name": "Mastodon",
+    "tos_url": "",
+    "url": "https://joinmastodon.org",
+    "wikipedia_url": "https://en.wikipedia.org/wiki/Mastodon_(software)",
+    "protocols": [
+      "OStatus",
```
strugee: We should probably add `ActivityPub` into this list; I expect more and more of the applications in the social category will be moving to support it since it's a W3C recommendation

outloudvi:
@strugee Tried to add one. Maybe there is something better to go of the self-host encouraging words.

yegortimoshenko:
Looks good to me. Thank you for this change, this is long overdue.

PR #1875, opened 2018-02-21
Switch to jsonlint2

strugee:
`jsonlint` is unmaintained and broken due to npm registry changes; see zaach/jsonlint#105. This will be failing all CI jobs that run in the near future, so we should fix it now.

I still need to update `package-lock.json` before this can be merged, self-assigning in an attempt not to forget.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "uglify-js": "^2.7.5"
   },
   "devDependencies": {
-    "jsonlint": "^1.6.2"
+    "jsonlint2": "^1.7.0"
   },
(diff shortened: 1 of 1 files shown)


strugee:
Oh I thought we had a lockfile for some reason. I'll merge as soon as this passes CI

alerque:
There is no `package-lock.json` even though there should have been. That being said I have an outstanding PR #1867 which should fix that by ditching `npm` support entirely in favor of `yarn` and adding the proper `yarn.lock` file to the repository.

PR #1876, opened 2018-03-01
Drop ixquick

yegortimoshenko:
Ixquick is defunct and now leads to Startpage (Google proxy). See: https://www.ixquick.com/


alerque:
Something doesn't scan here. The page you linked to is an instance of Startpage, but the thing you deleted is a different domain: https://ixquick.eu/ appears to be up and functioning as described in our listing and is not Startpage. I don't see any evidence of it being defunct here. What gives?

yegortimoshenko:
Interesting. Until recently, ixquick.com used to point to Ixquick, and now it points to Startpage. But yes, clearly ixquick.eu still points to Ixquick. I'll close this for now but I'll try to contact Ixquick to get a definite answer on whether the service is being deprecated or not.

yegortimoshenko:
I should also note that they removed all references to Ixquick from https://startpage.com. So it looks an awful lot like they're deprecating the service. I don't know PRISM Break's policy on that.

PR #1877, opened 2018-03-01
Mark YaCy as experimental

yegortimoshenko:
Unfortunately, YaCy is not ready for day-to-day use: search quality is very bad. For example, it can't find prism-break.org if I search for "prism break".

The concept is great but quality is so low that it perhaps warrants removal from the suggestions list according to "Quality over quantity" policy specified in PRISM Break's README file.

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -5687,7 +5687,7 @@
     "slug": "wordpress"
   },
   {
-    "development_stage": "released",
+    "development_stage": "experimental",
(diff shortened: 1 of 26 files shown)


strugee:
I too would be in favor of simply removing YaCy. It really feels like we're scraping the bottom of the barrel there. I'll file a different issue so we can cleanly separate that discussion.

yegortimoshenko:
Closing in favor of #1939.

PR #1878, opened 2018-03-01
Drop Seeks

yegortimoshenko:
Source code has been last updated in 2014, seeks-project.org is defunct, https://github.com/beniz/seeks/wiki/List-of-Web-Seeks-node doesn't point anywhere anymore, etc.

This is also related to discussion in #1123.


alerque:
You have an error in your URLs. Please be a little more careful doing research as this is the second time today your proposal is nullified by just correcting the URL: https://github.com/beniz/seeks/wiki/List-of-Web-Seeks-nodes

Admittedly lots of the nodes are down, but saying it points nowhere isn't true. I'm not sure this is quite as defunct as you propose. It does look like it is dying to me so this is still worth considering, but I think you overstated the case.

yegortimoshenko:
Thanks for noticing that! I'll try to be more careful.

My argument still stands: it is not OK to recommend experimental software that has not been touched by anyone for four years. There are only two working URLs in this list, and both have switched to searx (https://seeks.ru and https://github.com/beniz/seeks/wiki/List-of-Web-Seeks-nodes; see #1123). Given that there are no public Seeks nodes, I don't think that case is overstated after all.

alerque:
Again, a little more research before coming to conclusions might be in order. It appears the original repo got abandoned in favor [of this fork](https://github.com/asciimoo/searx) where the last commit was 8 days ago and has had 55 contributors to the original repository's 5. That's doesn't sound dead to me. It [looks like](http://stats.searx.oe5tpo.com/) there might be on the order of a hundred [currently active public nodes](https://github.com/asciimoo/searx/wiki/Searx-instances).

I'm going to suggest we update URLs to where the project is actually at these days rather than declaring its premature death.

alerque:
(Apologies for my own insufficiently researched message just now (deleted).)

It does look like this is a sinking or sunk ship and can be dropped.

yegortimoshenko:
I've rebased this against master, took a while to merge locales. Going ahead and merging this because Seeks is clearly abandonware at this point.

PR #1879, opened 2018-03-01
Remove Facebook from Email categories

yegortimoshenko:
While Facebook can conceivably used as an email service, it is not regarded as such and very few people use and/or know about this functionality. 

I think that putting Facebook in this category is confusing, hence this pull request.

### DIFF
--- a/source/db/en-projects-rejected.ls
+++ b/source/db/en-projects-rejected.ls
@@ -309,9 +309,9 @@ projects-rejected =
     name: "Facebook"
     categories: [
       * name: "Web Services"
-        subcategories: ["Email Accounts", "Social Networks", "Video & Voice"]
+        subcategories: ["Social Networks", "Video & Voice"]
       * name: "Servers"
-        subcategories: ["Email Server", "Social Networks", "Video & Voice"]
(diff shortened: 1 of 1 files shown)


alerque:
I think it should stay listed in that category.  Even if people don't think of it in terms of email, for many people it does serve in that capacity both in terms of messaging and in terms of a centralized ID on the web. Note the things we're listing there as replacements are not just email accounts either, some of them are communication suites. Not everything fits neatly into one category, which is why they can be listen under more than one.

yegortimoshenko:
> Note the things we're listing there as replacements are not just email accounts either, some of them are communication suites.

In Email Accounts category, Kolab Now, Autistici/Inventati and Riseup (listed replacements) all come with a proper email account. Every other proprietary service in that category (Gmail, Outlook, Yahoo! Mail, Yandex.Mail) also comes with a proper email account. Facebook is a clear outlier here.

Hillside502:
> messaging and in terms of a centralized ID

messaging = `Facebook Messenger`, not email

yegortimoshenko:
Merging because Facebook doesn't provide with an email account, so it doesn't belong to Email Accounts category.

PR #1880, opened 2018-03-01
Add OpenDNS to proprietary DNS services

yegortimoshenko:
Logo: https://commons.wikimedia.org/wiki/File:Opendns_logo.svg

The name is a misnomer and may lead to people believing that it's somehow more open than other DNS services. That is not the case. Cisco, current owner of OpenDNS, has helped Chinese government to build Great Firewall of China. OpenDNS privacy policy (available at https://www.opendns.com/privacy-policy/) currently forwards to Cisco's privacy policy, which allows them to share user's personal data with everyone.

Service is still widely suggested as Google Public DNS alternative, so it would be valuable having it listed on PRISM Break.


alerque:
:+1: This is probably worth listing just because it is so commonly recommended and not at all open. The Cisco ownership is also a new thing and the privacy policy didn't used to be that bad, so many recommendations pre-date the issues with the service. As a final anecdote their IP's are frequently hijacked in some countries.

yegortimoshenko:
Doesn't seem to be the case: there is only `en-projects-rejected.ls` and no comparable files for other locales, and also I've built a few other locales (eo, ru) and the OpenDNS (w/ icon) shows up in proprietary DNS services list.

yegortimoshenko:
I assume you're OK with this; @alerque is. Locales inherit list of rejected projects. Feel free to revert if you think this hasn't been discussed enough.

PR #1881, opened 2018-03-01
Drop GNU social

yegortimoshenko:
See: #1723, #1873

/cc @strugee


alerque:
I'm good with both this an Mastodon going through.

PR #1882, opened 2018-03-01
Remove "web service" badge from Up1

yegortimoshenko:
Up1 is not a centralized web service, but rather a self-hosted service. Currently there is a lot of inconsistency in usage of this badge, but it seems to me that the only meaningful distinction is to mark centralized web services.

Otherwise, all networking server-client software would have to be marked as "web service", because user will always use some particular instance of it (think of OpenVPN and multiple OpenVPN providers).

Notably, Kolab, Nextcloud, Riot are self-hosted services that currently do not have this badge. This pull request adds Up1 to the list.

Specifically compare Kolab (doesn't have "web service" badge) and Kolab Now (does have "web service" badge). I think this example captures the essence of when this badge is applicable.

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -5545,14 +5545,35 @@
     ],
     "categories": [
       {
-        "name": "Servers",
+        "name": "GNU/Linux",
         "subcategories": [
           "Media Publishing",
           "Productivity"
         ]
       },
       {
-        "name": "Web Services",
+        "name": "BSD",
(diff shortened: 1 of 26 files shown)


Zegnat:
Up1 used to have a publicly hosted instance people could use, back when it was added to the list. As this is gone [by their own admission](https://github.com/Upload/Up1#public-server) it seems right to dump the web service badge :+1:

PR #1883, opened 2018-03-01
Mark eCryptfs as audited

### LINKED ISSUE #1048: encfs = audited -> eCryptfs = audited
Without discussing the significance of an 10-hour audit, i wanted to mention that the guy who did the 10-hour audit of encfs (mentioned on the page) also did an 10-hour audit of eCryptfs (which is not mentioned on the page)!

See: https://defuse.ca/audits/ecryptfs.htm

yegortimoshenko:
Closes #1048.

### DIFF
--- a/source/db/en-projects.json
+++ b/source/db/en-projects.json
@@ -1200,7 +1200,15 @@
         ]
       }
     ],
-    "slug": "ecryptfs"
+    "slug": "ecryptfs",
+    "audit": {
+      "author": "Taylor Hornby",
+      "date": "2014-01-22",
+      "development_stage": "released",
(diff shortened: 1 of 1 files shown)


alerque:
:+1: for merge, at least for the EN locale. Did you check to see if other locales are pulling this data from EN if there is not a localized version? If it's not doing that then this is fine to merge, but we should also open an issue to fix it because they should be (I know this works for other project related strings, just not sure about these).

yegortimoshenko:
You are correct, it is inherited. I've dropped changes in all locales except EN.

alerque:
That's better, thanks. It's one thing to have a bunch on untranslated strings showing up on the built site, it's a whole 'nother thing to cram localization files full on English strings. It makes it really hard to determine what's been done and what needs doing down the road.

yegortimoshenko:
#1048 is marked as "accepting PRs", and LGTM, so merging.

PR #1714, opened 2017-03-25
Remove dukgo.com from zh_CN-projects.json

paperbag:

PR #1887, opened 2018-03-02
Remove dukgo.com from zh-CN locale

yegortimoshenko:
Supersedes #1714.

That pull request also includes translation update, which doesn't cleanly apply on top of current master. I might not be able to salvage this work because I don't know Chinese.

### DIFF
--- a/source/db/zh-CN-projects.json
+++ b/source/db/zh-CN-projects.json
@@ -1085,30 +1085,6 @@
   },
   {
     "development_stage": "released",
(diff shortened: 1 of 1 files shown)


alerque:
:+1: for merge, but this doesn't need to supersede 1714, only one commit of it. That PR has other changes and will just need to be rebased. I can probably salvage it if we have somebody to review, but go ahead with this and we'll worry about that later.

yegortimoshenko:
This only removes a single non-existent service, reviewed by @alerque, so merging.

PR #1891, opened 2018-03-02
Update RetroShare metadata

yegortimoshenko:
http://retroshare.org and http://retroshare.sourceforge.net are defunct.


alerque:
:+1: for merge from me.

PR #1892, opened 2018-03-02
Remove Ubuntu notice

### LINKED ISSUE #1471: Will Ubuntu/Derivatives be considered again after the 16.04 update?

yegortimoshenko:
Resolves #1471. Ubuntu does not come with spyware since 16.04, [which is the current LTS version](https://www.ubuntu.com/download/desktop), and new installs are unlikely to use older versions. That means the notice can be safely discarded.


alerque:
This does not resolve that issue. Just because a couple guys comment saying something should happen doesn't mean that's the concurrences.  See earlier in the thread such as [here](https://github.com/nylira/prism-break/issues/1471#issuecomment-167903593) for why we're still not really considering this. Ubuntu still is driven largely by corporate sponsorship rather than the developer community and has burned us several times. Most of us aren't up for trusting them again yet and no serious evidence that we should do so has been forwarded on that thread.

Because we are not listing them as a recommendation, we get a constant barrage of suggestions to include it. Since we don't otherwise have a rejected projects list yet, The warning notice is there as reminder that it has been considered. And I think for now it should stay.

alerque:
Yes, I could see rewriting it a little. It  should also link to a more relevant issue, the one it's linked to doesn't have the lastest or bulk of the discussion.

hasufell:
Yes, it always has been primarily a trust issue.

yegortimoshenko:
OK, this will need another iteration.

Issue #1349, opened 2015-06-20
Bookie.io is shutting down


vyp:
https://github.com/bookieio/Bookie/blob/4db38443a9435c3776037251f522324350fa84a2/README.rst links to the twitter account of this tweet: https://twitter.com/BookieBmarks/status/606388573852430337, so it does seem official.

Might be important to note that it's the bookie.io website instance that's shutting down and the software can still be self hosted. However, it's unclear whether the software will still be developed... cc @bookieio.

Question is, should it still be outright removed from the site?

vyp:
There are other public instances <https://github.com/bookieio/Bookie/wiki/Bookie-instances/_history>, so even though development has stalled, it still seems to be viable for self-hosting, at least for now. Meaning it probably shouldn't be taken down. If this changes in the future, we can reopen this or create a new issue/pr.

PR #1893, opened 2018-03-02
Drop Bookie

yegortimoshenko:
Relevant links:

* https://github.com/bookieio/Bookie
* https://twitter.com/BookieBmarks
* http://bmark.us

Source code has not been updated for three years, and primary instance has been shut down: https://twitter.com/BookieBmarks/status/606388573852430337

Official site is defunct.


alerque:
This can still be self hosted (which was the idea all along, the handful of public instances were never supposed to be the deal). Given the number of forks where people have fixed a bug or two to keep running themselves I wouldn't call this dead yet.

yegortimoshenko:
None of the forks seem to be active anymore (the most recent one I've found was last touched one year ago, and only migrates software to Python 3 without changing it in any practical way).

While it's true that unlike proprietary service, free self-hosted service cannot die and take all data with it, it can become obsolete quickly if there is no mindshare behind it. It's clear that if Bookie were to be proposed to be added to PRISM Break suggestions today, it would not be accepted due to little to no activity.

I think that if there was no comparable software in the list, it might have been OK to leave it, but there is [Shaarli](https://prism-break.org/en/projects/shaarli/) which is updated multiple times a week, and there is [Wallabag](https://prism-break.org/en/projects/wallabag/) which is very popular libre read-it-later service that is updated at least one every two weeks.

Since we don't have any scoring system, I think that by listing unmaintained Bookie, we're doing Shaarli and Wallabag and disservice, by presenting all options as equally viable. If one were to start a Bookie instance today, even if they would be able to use it for a year or two, they wouldn't be able to continue using it for much longer unless they are a programmer because there are no people to help them with the software (no bugfixes, no new features, ...).

Note: [Bookie](https://github.com/bookieio/Bookie) is a self-described delicio.us replacement, and [we describe Shaarli](https://prism-break.org/en/projects/shaarli/) as a "delicious clone".

alerque:
Okay I'll buy that. I hadn't compared with other options, and with other better self hosted solutions I'm game for dropping Bookie.

yegortimoshenko:
Related: #1349

PR #1895, opened 2018-03-02
Use Nextcloud logo from Nextcloud login page

yegortimoshenko:
Before:

screenshot_2018-03-02_19-24-36

After:

screenshot_2018-03-02_19-46-43


alerque:
I'm not sure this is an improvement. Notably the mobile device app icon is white on blue as in the original here. I don't see the white on blue variant to be a clear advantage here.

Zegnat:
I personally do not like to use alternative logos when an official logo has been published. In the case of Nextcloud, [the press page](https://nextcloud.com/press/) will let you download [the logo files (ZIP archive)](https://nextcloud.com/wp-content/themes/next/assets/files/Logo_files.zip). When you look at the files named “logo” all of them feature the word mark:

Screen shot showing 12 preview thumbnails of logo files from the press kit.

The blue-on-white logo is available there too, but is always specifically tagged as “white”. I am not sure if PRISM Break has a history of picking the logo that best fits on its own white background, if there isn’t already a history of this I would personally prefer to stick to the most basic version of the logo that Nextcloud provides, `nextcloud-logo.svg`:

Image: PNG rendition of the nextcloud-logo file in the press kit.

yegortimoshenko:
I think the blue-on-white logo is more attractive because it blends in with other logos better (most featured logos do not have a background), and word mark is too small to be readable and is duplicated by our own text. But, of course, it's highly subjective.

strugee:
Honestly I don't have a problem with the blue background, and indeed have a slight preference for it. But I don't really care if we switch as long as we stick to the official logos, which would mean the wordmark stays either way.

yegortimoshenko:
Got you, thanks all for feedback. Let's leave it as is.

strugee:
@yegortimoshenko thanks for suggesting it :-)

PR #1899, opened 2018-03-03
Various spelling/terminology fixes

yegortimoshenko:
(Feel free to cherry-pick those that you like.)


Hillside502:
Agreed, except for "nonfree"
https://www.thefreedictionary.com/nonfree

(FSF is wrong)

alerque:
Ditto, these all look good but the FSF is wrong about non-free and shouldn't try to meddle with normal English language usage rules. It's one thing to use made up names for projects, its another to try to pretend that grammar isn't a thing.

yegortimoshenko:
Agreed. Dropped that change.

source/locales/ar.json
```diff
@@ -168,7 +168,7 @@
   "app-store-notes": "--",
   "Meshnet": "Meshnet",
   "mesh-networks-project": "A decentralized alternative to the internet.",
-  "mesh-networks-notes": "<p>A mesh-networks is a decentralized peer-to-peer network, with user-controlled physical links that are usually wireless.</p><p>“Mesh networking (topology) is a type of networking where each node must not only capture and disseminate its own data, but also serve as a relay for other nodes, that is, it must collaborate to propagate the data in the network.“</p><p>—<a href='https://en.wikipedia.org/wiki/Mesh_networking'>Wikipedia</a></p>",
+  "mesh-networks-notes": "<p>Mesh network is a decentralized peer-to-peer network, with user-controlled physical links that are usually wireless.</p><p>“Mesh networking (topology) is a type of networking where each node must not only capture and disseminate its own data, but also serve as a relay for other nodes, that is, it must collaborate to propagate the data in the network.“</p><p>—<a href='https://en.wikipedia.org/wiki/Mesh_networking'>Wikipedia</a></p>",
```
strugee: All these should be "A mesh network" instead of "Mesh network"

source/locales/ja.json
```diff
@@ -145,7 +145,7 @@
...
   "Scheduling": "スケジュール",
   "Online Office Suite": "オンラインオフィススイート",
-  "productivity-notes": "<p><strong>etherpad</strong> プロジェクトが <a href='https://github.com/ether/etherpad-lite/wiki/Sites-that-run-Etherpad-Lite'>etherpad サービスを実行するサイト</a> の一覧を維持しています。 信頼してデータを扱う前に、SSL を使用するサービスだけを選択して、サイトの背景を調査してください。</p><p><a href='#email-accounts'><strong>Riseup</strong></a> もメール、XMPP、チャット、(Up1 経由) データホスティングサービスを提供しています。 すべて <a href='#mesh-networks'>Tor 秘匿サービス</a> アドレスを通じてアクセスすることができます。 これらのアドレスの一覧は <a href='https://www.riseup.net/en/tor#riseups-tor-hidden-services'>ここで利用可能です</a>。</p>",
+  "productivity-notes": "<p><strong>Etherpad</strong> プロジェクトが <a href='https://github.com/ether/Etherpad-lite/wiki/Sites-that-run-Etherpad-Lite'>Etherpad サービスを実行するサイト</a> の一覧を維持しています。 信頼してデータを扱う前に、SSL を使用するサービスだけを選択して、サイトの背景を調査してください。</p><p><a href='#email-accounts'><strong>Riseup</strong></a> もメール、XMPP、チャット、(Up1 経由) データホスティングサービスを提供しています。 すべて <a href='#mesh-networks'>Tor 秘匿サービス</a> アドレスを通じてアクセスすることができます。 これらのアドレスの一覧は <a href='https://www.riseup.net/en/tor#riseups-tor-hidden-services'>ここで利用可能です</a>。</p>",
```
strugee: This URL change seems to have been done accidentally

source/db/nl-projects.json
```diff
@@ -4701,7 +4701,7 @@
     "description": "Veilige communicatiemiddelen voor activisten.",
     "license_url": "",
     "logo": "riseup.png",
-    "notes": "Riseup's services may also be accessed via their Tor Hidden Service addresses. A list is [available here](https://www.riseup.net/en/tor#riseups-tor-hidden-services).",
```
strugee: Did you check if there are any instances of "hidden services"/"Hidden Services", plural?

source/locales/nl.json
```diff
@@ -163,7 +163,7 @@
   "DNS Provider": "DNS-provider",
   "dns-notes": "<p>Google Public DNS logt je ISP en locatiegegevens permanent voor analyse. Bovendien wordt je IP-adres voor 24 uur opgeslagen.</p><p><strong>OpenNIC</strong> heeft geen officiële policy aangaande log query privacy / anonimisatie. Meer informatie in het <a href='https://wiki.opennic.org/opennic:tier2'>OpenNIC-wiki</a>.</p>",
   "Anonymizing Networks": "Anonieme netwerken",
-  "Tor Hidden Services": "Tor-verborgen Services",
+  "Tor onion services": "Tor-verborgen Services",
```
strugee: Are these keys expected somewhere else in the codebase? I haven't checked at _all_ but it seems like this might wreak havoc

strugee: Also the translations are presumably wrong too, we should probably get rid of them

yegortimoshenko: I'm not sure, while Tor hidden service is a deprecated name for this technology, I think it's better to use it than have no translation at all.

source/locales/ja.json
yegortimoshenko: True! I'll fix that.

(However, GitHub route matching is not case-sensitive, so it doesn't matter much.)

yegortimoshenko:
Thanks for the review. I've extracted problematic change (hidden service -> onion service) into #1944, and integrated your feedback concerning other changes.

yegortimoshenko:
OK, I've double checked changes and looks good.

PR #1903, opened 2018-03-04
Remove mirrors (defunct)

yegortimoshenko:
Remove defunct mirrors. Unfortunately, that includes all mirrors listed in `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,13 +127,4 @@ Remember to revert the `Makefile` change and then commit the changes and issue a
 
 If you wish to mirror this site, [nylira/prism-break-static](https://github.com/nylira/prism-break-static) is probably of interest to you. This is a completely static (but constantly updated) version of the site you can save to browse locally or serve over HTTP.
 
-### Mirrors
-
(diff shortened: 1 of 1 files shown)


yegortimoshenko:
I'm not sure nylira/prism-break-static is a good idea. Could we perhaps set up Travis CI to build every commit into a release archive that can be unpacked and served? See https://github.com/yegortimoshenko/zhenya-slim-theme/blob/master/.travis.yml for example.

strugee:
> I'm not sure nylira/prism-break-static is a good idea.

Can you elaborate? Why? I think it's much better than publishing archives because you can just put `git pull` on a cronjob.

FYI nylira/prism-break-static is what is served on production PRISM Break. IIRC @nylira runs it every day so it takes max 24h for changes to go live.

yegortimoshenko:
Compare trees, I don't think @nylira runs it every 24h:

https://github.com/nylira/prism-break/commits/master
https://github.com/nylira/prism-break-static/commits/master

Either way, it just seems to be sort of work that could be automated. Cron job is a valid concern, though.

PR #1905, opened 2018-03-04
Remove notice about jsonlint from CONTRIBUTING.md

yegortimoshenko:
`make test` runs jsonlint, so no need to repeat that step. Also, we now use jsonlint2 (60f7039fcc68057f11c5639701f0e46f92506255).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,7 +99,6 @@ If you're interested in translating the site itself (all the nouns, verbs, and s
 If your language file doesn't exist yet, you can copy the en.json file and start translating from there.
 
 ### JSON Validation
-**Make sure your JSON validates.** You can use either use [JSONLint](http://jsonlint.com/) online or install it locally with `npm install jsonlint -g`.
 
 A common mistake is putting unescaped quotation marks in a sentence. Make sure to use Unicode typographic marks (curly quotes) or escape regular (straight) quotes with a backslash.
 


CONTRIBUTING.md
```diff
@@ -99,7 +99,6 @@ If you're interested in translating the site itself (all the nouns, verbs, and s
 If your language file doesn't exist yet, you can copy the en.json file and start translating from there.
 
 ### JSON Validation
-**Make sure your JSON validates.** You can use either use [JSONLint](http://jsonlint.com/) online or install it locally with `npm install jsonlint -g`.
```
strugee: I would prefer if we kept this note in some form. IIRC `make test` does more than just lint stuff (doesn't it do a full build?) and anyway it's a better contributor experience to catch errors locally rather than in CI.

Honestly, it seems okay the way it is.

strugee:
Also FWIW jsonlint2 and jsonlint are basically the same thing. We switched to jsonlint2 and then the author of jsonlint actually showed up to fix it, which I wasn't expecting. 60f7039fcc68057f11c5639701f0e46f92506255 could be reverted honestly but I didn't see the point in bothering. Feel free to send a PR.

CONTRIBUTING.md
alerque: I agree this could be kept, I would reword it to `make lint`. We don't have to explain _how_ it is validated or send people of to a website to copy their code into because we have that as part of the build process they can run themselves. This whole section should probably just be "Validation" or "Test" without mentioning JSON in particular and encourage them to test locally before shipping it off to CI.

strugee: @alerque definitely agree. FWIW I do see some value in this:

> send people of to a website

being kept, since some people want to casually contribute without setting up a full Node/yarn environment. Maybe "just see what CI says" is good enough for this segment though?

alerque: @strugee Yes, simple is fine.

strugee:
So, I'm closing this based on https://github.com/nylira/prism-break/pull/1905#discussion_r172442230. Feel free to send another PR :thumbsup:

PR #1908, opened 2018-03-06
Add a Travis CI badge to README.md

strugee:


strugee:
If anyone thinks these are ugly speak now or forever hold your peace

(Or at least until you file a new issue)

yegortimoshenko:
I don't really care either way, so feel free to merge, but on a more general tangent, it's incredible that Git's decentralized promise has never manifested itself, as of date, except in the kernel itself. 

(It would not make sense to add Travis CI build status badge to Linux kernel, because there's no authoritative copy to build. Of course, such badges necessarily bless one of the versions as the authoritative version, which leads to centralization. And Travis CI is proprietary.)

strugee:
> And Travis CI is proprietary.

[No it's not](https://github.com/travis-ci).

I'm just going to skip over the remark about Git's decentralization since I don't want to get into an argument I don't care about.

yegortimoshenko:
> No it's not.

You seem to be correct, from Wikipedia:

> Although the source is technically free software and available piecemeal on GitHub under permissive licenses, the company notes that it is unlikely that casual users could successfully integrate it on their own platforms.

It's just there's no way to self-host it, so I thought they only liberate certain components.

PR #1919, opened 2018-03-11
Adopt Ruby code of conduct

yegortimoshenko:
See: https://www.ruby-lang.org/en/conduct/

This is an alternative to #1790. The proposed code of conduct is used by Ruby interpreter and is similar in spirit to [Debian code of conduct](https://www.debian.org/code_of_conduct) and [Fedora code of conduct](https://docs.fedoraproject.org/fedora-project/project/code-of-conduct.html).

I think this code of conduct would be more approachable and readable (something we want so that people who participate actually read it) and also significantly less political/inflammatory than Open Code of Conduct that is currently being used.

This commit also moves Code of Conduct section from the very bottom to the very top of `CONTRIBUTING.md` so that more people will notice it.

/cc @alerque @hasufell @strugee @vyp

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,10 @@
+# Code of Conduct
+
+* Participants will be tolerant of opposing views.
+* Participants must ensure that their language and actions are free of personal attacks and disparaging personal remarks.
+* When interpreting the words and actions of others, participants should always assume good intentions.
+* Behaviour which can be reasonably considered harassment will not be tolerated.
+
 # Contribute: Projects
 
 ### Which Files to Edit
@@ -130,9 +137,3 @@ If you wish to mirror this site, [nylira/prism-break-static](https://github.com/
 
  - [https://www.sedrubal.de/service/prism-break/en/](https://www.sedrubal.de/service/prism-break/en/)
  (credit: [sedrubal](https://github.com/sedrubal))
-
-# Code of Conduct
-
-This project adheres to the [Open Code of Conduct][code-of-conduct]. By participating, you are expected to honor this code.
-
-[code-of-conduct]: http://todogroup.org/opencodeofconduct/#PRISM-Break/abuse@prism-break.org


strugee:
I'm -1 on this.

I would much rather have a stronger CoC that explicitly lists -isms that we don't tolerate. This is important in any community but especially in PRISM Break because the marginalized communities that stronger CoCs protect are often the people who need resources like PRISM Break the most. Having a stronger CoC is, among other things, an explicit signal that people who experience e.g. sexism and racism are welcome here and won't be attacked. CoCs like the Ruby one are too open-ended and vague for this signaling to be effective.

yegortimoshenko:
> Having a stronger CoC is, among other things, an explicit signal that people who experience e.g. sexism and racism are welcome here and won't be attacked. 

I'm concerned that people are too aggressive at times, not as much due to discrimination, but due to the fact that this is a very opinionated list that excludes projects more often that the opposite. It tends to cause conflict by its very nature, and it would be nice to reduce the rate of that happening.

It is my belief that all -isms can't be listed, everyone has their struggle, and everyone has to treat each other well, no matter if particular discrimination is in that list or not. It doesn't seem reasonable, or even possible, to list every and each one, instead of extracting the general principle of treating other people well.

Some might also not like unfriendly displays of authority and power (like "may face temporary or permanent repercussions", "ban temporarily or permanently any contributor"). Even if they all are true, they increase the temperature (see https://freenode.net/changuide) and are couter-productive to more peaceful communication that I would like to achieve.

That said, I agree that Ruby code of conduct might be too vague and not reinforcing. I very much like Debian's [code of conduct](https://www.debian.org/code_of_conduct) and [diversity statement](https://www.debian.org/intro/diversity), because both strike at heart of the problem. Do you think it's any better?

strugee:
> I'm concerned that people are too aggressive at times, not as much due to discrimination, but due to the fact that this is a very opinionated list that excludes projects more often that the opposite.

I agree. I definitely would like this to be addressed in any CoC we adopt; it's been long enough that I don't know how well #1790 does this.

> It is my belief that all -isms can't be listed

You can't list all of them but it helps to call out as many as possible; this is why most such lists start with "including but not limited to"

> Some might also not like unfriendly displays of authority and power (like "may face temporary or permanent repercussions", "ban temporarily or permanently any contributor"). Even if they all are true

What do you mean by "true" here? Even if what are all true? In any case I don't see how to avoid this. It _has_ to be possible to ban people from a community if they're toxic for it. Otherwise what use is having a code of conduct?

yegortimoshenko:
> You can't list all of them but it helps to call out as many as possible; this is why most such lists start with "including but not limited to"

That in itself places one struggles over the others, leaving people fighting the latter feel excluded.

> What do you mean by "true" here? Even if what are all true? In any case I don't see how to avoid this. It has to be possible to ban people from a community if they're toxic for it. Otherwise what use is having a code of conduct?

Sorry, that's not what I meant. My point is that threatening people is not very helpful, because it often makes people aggressive towards the one issuing threats, even if those threats are for just and good causes, and even when threats were not towards people that read it.

Notice how Debian code of conduct and diversity statement never threatens: only appeals to ethical values and human nature that we share.

hasufell:
> I would much rather have a stronger CoC that explicitly lists -isms that we don't tolerate.

Definitely no to this.

What you are establishing with that is simply political correctness, which means nothing. You want to establish a **culture**, so describe that culture instead of listing "no-gos".

PR #1921, opened 2018-03-11
Rephrase Iceweasel notice, resolves #1559

### LINKED ISSUE #1559: Consider removing mention of Iceweasel now
[Firefox resumes distribution in Debian](https://wiki.debian.org/Iceweasel#Firefox_resumes_distribution_in_Debian)

"Firefox is now in Debian again with the official name and branding...iceweasel is a transitional package which installs the firefox-esr package"

yegortimoshenko:
Resolves #1559. Before:

> Another debranded alternative to Firefox is <a href='https://wiki.debian.org/Iceweasel'>Iceweasel</a>, a browser for Debian-based distributions.</p>

After: 

> Iceweasel is a rebranded version of Firefox that does not have to follow <a href="https://www.mozilla.org/en-US/foundation/trademarks/policy/">Mozilla Trademark Policy</a>.


strugee:
I am not bothered by Firefox's trademark policy. It has absolutely no impact on the reasons we require PRISM Break listings to be FLOSS, which is mainly auditability and secondarily the fact that the threat of a fork tends to deter antifeatures. Practically speaking the trademark issue really does not have an effect on software freedom. It's just nitpicking.

I'm sure there's an argument to be made about intellectual property laws, and how they're messed up, but I don't think PRISM Break should have opinions on issues like that.

yegortimoshenko:
I think you misread this issue. The point here is that people know that Iceweasel and Firefox are the same thing, not trademark issue.

I'd of course argue on that token, I think you underestimate importance of being able to make changes to source code and share in a way that is recognizable to other people. Owning words is a source of incredible power and thus control. But this issue is not about that.

yegortimoshenko:
Merging this because there's no anti-IP sentiment behind the new wording, just making sure Parabola/GuixSD/... folks know that Iceweasel and Firefox is the same thing.

PR #1923, opened 2018-03-13
Update Metronome metadata

yater:
broken link replaced (only for the english content (TODO must be changed for all languages))


yegortimoshenko:
Thanks. I will make this change in all files on merge.

strugee:
@yegortimoshenko can you instead push a commit to this branch so we can all review?

yegortimoshenko:
That's what I was going to do! :-)

yegortimoshenko:
I've merged this patch as part of #1929. Thanks!

PR #1939, opened 2018-04-07
Drop YaCy

### LINKED ISSUE #1888: Consider removing YaCy
See https://github.com/nylira/prism-break/pull/1877#issuecomment-369790653; YaCy's quality is extremely low. Admittedly I last tried it several years ago but I remember it being clunky, Java-based and not terribly useful overall. I really feel like we're scraping the bottom of the barrel with this one and I'd rather just remove it.

yegortimoshenko:
Resolves #1888.


yegortimoshenko:
OK, I've tried building a few locales, LGTM. Merging.

PR #1952, opened 2018-04-07
Add Tinfoil Chat

yegortimoshenko:
This is a follow-up to #1178, with resolved conflict. This should be considered for merge when TFC implements an onion service backend, and when @maqp feels it's time.

screenshot_2018-04-07_21-01-50


maqp:
@yegortimoshenko et. al. Development of FOSS systems is not free. Having paid everything from my own pocket for years, it's unfortunate and understandable I would eventually need to put the development of TFC on hold. At current time I have no estimate how soon I'm able to continue. But rest assured, it's not due to lack of motivation. Meanwhile, everything from Python as a language, the supporting software, used operating systems, cryptographic libraries and of course, Tor and v3 Onion Services, will mature.

Issue #1936, opened 2018-04-04
Delete Riot (Instant Messaging and Video & Voice)


RiQuY:
The open source idea and decentraliced network was a good idea, but just take a look at the Privacy Policy. Riot will store everything you send and that's pretty bad and intrusive.

https://riot.im/privacy
Read the part "What kinds of information do we collect?":
Also check "When might we disclose your information?"

Zegnat:
Not sure how to handle this. First of all it looks like a pretty standard boilerplate privacy policy for any sort of online service. And service is the key word there.

The Riot _Service_ is what you use when you do not run the server software yourself. Of course this service will have access to the data you are sending by using it. Of course this data may contain “personally identifiable data”, and of course you explicitly give it PID when you tell the service that you want to to authenticate to Slack (and others) on your behalf.

Note that they also make it clear that the communication content “may be encrypted by your client”. That’s how you make sure the service cannot know what you are sending, by using encryption on your end. According to [the security note](https://prism-break.org/en/projects/riot/) Riot themselves are working on putting end-to-end encryption in their web client, actively working to make sure they will not accidentally end up owning your data.

I am split on this issue. There are enough other IM options listed, and I wouldn’t see removing Riot as the biggest loss. On the other hand, the argument made here is to throw out everything that offers a hosted version.

What about Diaspora? We haven’t checked the privacy policies on all the recommended Pods. It should be implied that PRISM Break doesn’t recommend all Pods and encourages you to self host. The same applies to Riot.

RiQuY:
I think that the worst part is that he can sell our data to search engines or others, as seen in "When might we disclose your information?".
I know that the server maybe needs to maintain the data, but they should not share it in any circunstance.

Zegnat:
You mean this line?

> We may share personal information with analytics and search engine providers that assist us in the improvement and optimisation of our Website and Service.

This is why I expect this to be some sort of boilerplate privacy policy. Things like your IP address and browser information are PID depending on jurisdiction. And obviously those sort of statistics are being recorded. Rather than trying to specify what info they collect for analytics and what they do not, they rather just put a blanket statement on there.

It might be worth contacting them to clear that up: do they even use *external* analytics services? What *exactly* do they hand over?

I’ll be interested in seeing what other people think.

Hillside502:
Riot the service hides behind Cloudflare
https://iplookup.flagfox.net/?ip=104.27.178.139&host=riot.im
[...]

PR #1953, opened 2018-04-07
Drop Riot from IM and Email Alternatives categories

yegortimoshenko:
We should not recommend software that is not end-to-end encrypted by default for instant messaging (or any private one-to-one messaging, for that matter). Riot might still be fine as an alternative to IRC or Slack.

This commit also removes duplicated non-localized Riot entry from all locales except `en`.

See: https://github.com/nylira/prism-break/pull/1178#issuecomment-379494346

> We already have Signal, Ricochet etc that provide strong security on networked endpoints. Yet Prism-Break takes time to feature "experimental" apps like Riot that do not provide end-to-end encryption by default. I mean, Facebook Messenger, Google Allo and Telegram all sit in the same boat, yet the site shuns away from those.

There's also #1936.


ara4n:
Matrix (and therefore Riot) is one of the only decentralised E2E-encryption capable comms systems there is.  The reason to include it on prism-break is that it is **decentralised** (as well as E2E capable), unlike Facebook Messenger, Google Allo & Telegram, and so users have the freedom to use whatever server they like (even if they are confused by the privacy policy on the default matrix.org server).

We haven't turned on E2E by default because it is still in beta and we want to ensure there are *no* regressions in UX when enabled, but we make it *very* clear when it's turned off:

screen shot 2018-04-26 at 15 22 11

...and it's then trivial to turn on when needed.  At the current rate we hope to have it enabled by default at some point in the next ~3 months (although we are juggling scaling challenges at the same time).

So, I think that it's still relevant for IM & email purposes.

yegortimoshenko:
> Matrix (and therefore Riot) is one of the only decentralised E2E-encryption capable comms systems there is. 

This is true; the only other one is XMPP.

> ...and it's then trivial to turn on when needed. At the current rate we hope to have it enabled by default at some point in the next ~3 months (although we are juggling scaling challenges at the same time).

Great! Then there is no need for this. Thank you a lot for a heads up :-)

I am closing this in favor of https://github.com/nylira/prism-break/issues/1963 which just tracks E2E progress.

PR #1956, opened 2018-04-08
Drop GNUnet

yegortimoshenko:
The project is practically dead. The latest release dates back to 2014, there has been zero activity in the repositories in the last 14 months: https://gnunet.org/git/

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -1865,54 +1865,6 @@
   },
   {
     "development_stage": "released",
-    "description": "GNUnet is a fully free P2P network.",
-    "license_url": "https://gnunet.org/svn/gnunet/COPYING",
-    "logo": "gnunet.png",
-    "privacy_url": "",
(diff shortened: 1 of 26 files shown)


yegortimoshenko:
I would like to go ahead with this unless anyone minds.

strixaluco:
I see that GNUnet projects repos are more than just alive: https://gnunet.org/git/?s=idle
Note the ```Idle``` column.
[Main repo](https://gnunet.org/git/gnunet.git/log/) reveals dozens of commits daily.

yegortimoshenko:
@strixaluco: Hm, why https://gnunet.org/git/ doesn't show those repos? This is the URL everyone links to.

strixaluco:
It does, but it links to umbrella project. We can change the URL to `https://gnunet.org/git/gnunet.git/log/` at least at PRISM Break.
By the way, https://prism-break.org/ seems to be down ATM.

PR #1961, opened 2018-04-22
Translate "Onion services" at zh-cn

outloudvi:
Putting a translation there in case of a empty bar seeming weird.
Maybe not a really good translation. Suggestion welcomed.


yegortimoshenko:
Official translation is 洋葱服务, see: https://tb-manual.torproject.org/zh-CN/onion-services.html

yegortimoshenko:
Thanks for the contribution!

Issue #1975, opened 2018-05-30
Remove Ricochet from "Encrypted Instant Messenger"


ohmynameisrico:
https://prism-break.org/en/categories/windows/#instant-messaging

https://github.com/ricochet-im/ricochet/issues/579

ohmynameisrico:
https://trac.torproject.org/projects/tor/wiki/doc/TorifyHOWTO/InstantMessaging

Copy start:
---

Software listed here has become depreciated and development has ceased. They are therefore considered unsafe and it is highly recommended to avoid using them. This section is here purely for historical value.

Ricochet IM is an open-source, decentralised instant messenger project that officially ended development in November 2016. Ricochet starts a Tor onion service on the users local system and facilitates communication with other Ricochet users whom are also running their own Ricochet-created Tor onion service, providing End-to-End encryption by never allowing the connection to leave the Tor network.

---
Copy End.

strugee:
So, I too have noticed Ricochet has not had a release in a long time. I would like to look into this.

However we will need something more concrete than this. All I see in the linked issues is FUD and telling people to reread the FUD. For example, a useful question to answer would be: is the Tor shipped with Ricochet really out of date? Or is it fine because it's a Tor LTS release?

yegortimoshenko:
Ricochet ships 0.2.8.9:

```
$ tar jxf ricochet-1.1.4-linux-x86_64.tar.bz2
$ ricochet/tor --version
Tor version 0.2.8.9.
```

See this list of Tor vulnerabilities: https://trac.torproject.org/projects/tor/wiki/TROVE

Ricochet is vulnerable to every vulnerability in the list except the first one. I think we should pull it down ASAP.

ohmynameisrico:
@strugee It's not FUD. And stop saying FUD because I hate that word.
Ricochet is vulnerable.

Zegnat:
> Tor <i>[version 0.2.8.9]</i> bundled with Ricochet is vulnerable to every vulnerability in the list except the first one.

Reading https://github.com/ricochet-im/ricochet/issues/578#issuecomment-393049811 it sounds like packaged versions of Ricochet will actually ship with newer versions of Tor? Or prefer using whatever version of Tor is locally installed?

@yegortimoshenko are you currently at a machine where you can verify that?

> It’s not FUD. And stop saying FUD because I hate that word.

[...]

PR #1977, opened 2018-05-30
Remove Ricochet from "Encrypted Instant Messenger"

ohmynameisrico:
https://github.com/nylira/prism-break/issues/1975


ohmynameisrico:
https://github.com/ricochet-im/ricochet/issues/579

yegortimoshenko:
5c053bcf14c10d2adc6399d54b8c028de6f57b6b does the same, but removes Ricochet from all locales.

PR #1981, opened 2018-06-01
Drop Tent

yegortimoshenko:
This Tent thing has seemingly never took off, all of its repos have been last updated at 2016: https://github.com/tent


strixaluco:
Apparently the team switched to work on Flynn instead: https://github.com/flynn

PR #1987, opened 2018-06-01
Drop DAVDroid in favor of Etesync

yegortimoshenko:
We already list Etesync, which does the same thing as Davdroid while
being end-to-end encrypted, as in not leaking any data to the server.

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -82,12 +82,6 @@
     "slug": "cryptsetup"
   },
   {
-    "description": "DAVdroid is an open-source CalDAV/CardDAV synchronization app for Android 4+.",
-    "url": "http://davdroid.bitfire.at/",
-    "wikipedia_url": "",
-    "slug": "davdroid"
-  },
-  {
     "description": "توزيعة جنو/لينكس شهيرة و أخلاقية.",
     "url": "http://debian.org",
     "wikipedia_url": "https://en.wikipedia.org/wiki/Debian",
--- a/source/db/ca-projects.json
+++ b/source/db/ca-projects.json
@@ -82,12 +82,6 @@
     "slug": "cryptsetup"
   },
   {
-    "description": "DAVdroid is an open-source CalDAV/CardDAV synchronization app for Android 4+.",
-    "url": "http://davdroid.bitfire.at/",
-    "wikipedia_url": "",
-    "slug": "davdroid"
-  },
-  {
     "description": "Popular distribució de GNU/Linux famosa pel seu contracte social.",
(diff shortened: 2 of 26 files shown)


Zegnat:
I seem to vaguely recall DAVDroid could be pinned to a single certificate, which makes it end-to-end encrypted as well. I believe there is still definitely a use-case to be made for *DAV based applications, as many server infrastructures will give people that option.

yegortimoshenko:
You can pin down a certificate, that would resolve possibility of PKI MITM, but server still knows all your contacts and calendars, these are stored in plain text. No knowledge storage fundamentally solves the trust problem, by people not having to trust anyone with their data.

yegortimoshenko:
Etesync uses CalDAV/CardDAV, too.

Zegnat:
I think listing an application that can be used to communicate with a *DAV server as securely as possible has value, especially as the server side may not be under ones own control (e.g. company managed).

> Etesync uses CalDAV/CardDAV, too.

If the Etesync app can replace DAVDroid for non-Eyesync servers without losing usability. That seems fine to me. I couldn’t figure that out with a quick search, only running into a server-proxy to make other *DAV applications able to query an Etesync server.

yegortimoshenko:
Yes, server implementation is not compatible, so one can't just use, say, Google's CardDAV server. But clients can use Etesync's DAV proxy implementation alongside with Etesync server.

I think a critique of yours applies here: are proprietary CardDAV implementations less privacy-respecting than libre DAVDroid? I think they are just as prviacy-respecting, so there's little value in listing DAVDroid specifically. Etesync is an alternative that is much more privacy-respecting, and doesn't have any disadvantages over DAVDroid other than its popularity: its Android client is even built on top of DAVDroid.

rfc2822:
Just wanted to note that CalDAV/CardDAV cannot be "end-to-end encrypted" by design. The protocols required that the server understands (and transforms etc.) contact/event data. It is possible to use WebDAV to store encrypted contacts/events on the server, but that's not CalDAV/CardDAV. For instance, a CalDAV server MUST be able to send a list of events within a certain time range. Such things would be impossible if the server doesn't have event data.

So DAVdroid will never be end-to-end encrypted, because it has another trust model (the one by CalDAV/CardDAV). If you need full privacy with CalDAV/CardDAV, you need a trusted server.

yegortimoshenko:
@rfc2822 Sure, that's why Etesync is neat: it emulates Cal/CardDAV via a localhost proxy and uses a different protocol behind the scenes which both facilitates a journaled storage and zero knowledge on server's side.

PR #1989, opened 2018-06-01
Drop Syndie

yegortimoshenko:
Hasn't been updated since 2016, see:
https://github.com/i2p/i2p.syndie/commits/master

Also, only works with I2P.


JeremyRand:
> Hasn't been updated since 2016

More precisely, the only 2 updates since then were a documentation fix and a Debian packaging fix; not related to the core code.

PR #1990, opened 2018-06-01
Drop Openswan

yegortimoshenko:
Openswan has not been updated for the most part since 2017:
https://github.com/xelerance/Openswan/commits/master

Strongswan does the same thing but is way better maintained:
https://github.com/strongswan/strongswan

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -593,12 +593,6 @@
     "slug": "openstreetmap"
   },
   {
-    "description": "Free software VPN client derived from FreeS/WAN.",
-    "url": "https://www.openswan.org/projects/openswan/",
-    "wikipedia_url": "https://en.wikipedia.org/wiki/Openswan",
-    "slug": "openswan"
-  },
-  {
     "description": "Free software VPN client.",
     "url": "http://openvpn.net/index.php/open-source.html",
     "wikipedia_url": "https://en.wikipedia.org/wiki/OpenVPN",
--- a/source/db/ca-projects.json
+++ b/source/db/ca-projects.json
@@ -593,12 +593,6 @@
     "slug": "openstreetmap"
   },
   {
-    "description": "Free software VPN client derived from FreeS/WAN.",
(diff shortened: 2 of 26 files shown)


Zegnat:
If we’re saying Openswan is bad because Strongswan is a thing, I would rather see a straight-up replacement than just dropping Openswan.

Also note that both [the debian specific fork](https://github.com/xelerance/Openswan/tree/debian-sid) and [the current development fork](https://github.com/xelerance/Openswan/tree/2.6.51dev), have more recent updates. In fact, commits seem to be coming in at a steady pace throughout the last months on the development fork.

yegortimoshenko:
Actually I think we should drop IKE VPN servers entirely in favor of OpenVPN and Wireguard because there are major concerns over the protocol spec's origin.

yegortimoshenko:
Even if we were to list an IKE VPN implementation, since we're not a free software registry, it would be nice to just list one. The one that gets the most of community's attention, is the most active and has most documentation currently is Strongswan.

Zegnat:
I am not using IPSec and I was merely commenting based on what you wrote in the original issue.

I am not against nudging people away from IPSec and onto OpenVPN and Wireguard, but slipping that debate in on a PR about dropping Openswan feels a little disingenuous. (Maybe?) And sounds like it may be worth its own issue. If others also agree that there is no security/privacy reason to keep IPSec around, I am all for dropping Openswan without replacement!

yegortimoshenko:
I guess what I meant is, I would certainly like to have discussion to drop IPsec entirely, but short-term I would at least like to drop Openswan because it's an actively harmful advice. There are way, way better maintained forks these days, we're listing three (Openswan, Libreswan, Strongswan).

Zegnat:
For some reason I didn’t notice Strongswan was already listed 😲 Then my original objection is out the window. This PR doesn’t need to replace Openswan with Strongswan because Strongswan is already listed.

No objects 👍

yegortimoshenko:
Yes, sorry I haven't communicated that clearly from the beginning.

PR #1992, opened 2018-06-01
Drop Disconnect

yegortimoshenko:
Disconnect comes with nonfree software: https://github.com/disconnectme/disconnect/issues/366

Repository has seen very little activity since 2015 (4 commits total): https://github.com/disconnectme/disconnect/commits/master

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -107,12 +107,6 @@
     "slug": "diaspora"
   },
   {
-    "description": " توقف مواقع خارجية من تتبعك.",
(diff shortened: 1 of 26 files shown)


yegortimoshenko:
cc @davidhedlund

Zegnat:
I believe the Disconnect blocking list has been adapted and is supported by a number of actually free blockers now? I see no real reason why we need to recommend Disconnect alongside other blockers at this point.

yegortimoshenko:
@Zegnat I believe so, too.

davidhedlund:
Thanks.

PR #1996, opened 2018-06-01
Remove JonDo Live CD and JonDoFox

yegortimoshenko:
Jondo mix operators log if asked by court order, that functionality is implemented in mix node source code: https://anonymous-proxy-servers.net/en/law_enforcement.html, https://anon.inf.tu-dresden.de/strafverfolgung/policy_en.pdf

This is further complicated by the fact that free cascades only have two hops: https://anonymous-proxy-servers.net/en/status/

Premium cascade (the only feature that could justify listing JonDoFox, 3 hops) means that all traffic of any specific user is filed under single profile (read bottom of https://anonymous-proxy-servers.net/en/premium.html) and essentially nothing stops mix operators from logging any specific users' traffic if they wish so (as long as they are an exit node).

Also notice that order of cascades is fixed (see status page above), so there are only 11 exit nodes in premium cascade network total, governed just by 4 operators total (!), while in free cascade network there are just two exit nodes both governed by the same operator.

It's not clear that JonDoFox keeps up to date security-wise: see https://jondobrowser.jondos.de/releases/current/, last updated 5 months ago. Also, relative obscurity of this software means we don't really know if it meets security characteristics it describes.

I think it's very bad that we mislead users into believing that JonDoFox is comparable to Tor browser in any way or form.

JonDo Live CD is even worse, and hasn't been updated since 2016-01: https://anonymous-proxy-servers.net/en/jondo-live-cd.html


alerque:
Agreed, that recommendation needs to be scratched entirely for now.

PR #2004, opened 2018-06-02
Remove Etherpad

yegortimoshenko:
Now that we list a comparable zero-knowledge service (Cryptpad), we shouldn't encourage Etherpad because it requires trusting server with plain text data.

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -1241,37 +1241,6 @@
   },
   {
     "development_stage": "released",
-    "description": "وثائق  تعاونيّة،  مفتوحة  المصدر في  الوقت  الحقيقي.",
-    "license_url": "https://github.com/ether/etherpad-lite/blob/develop/LICENSE",
-    "logo": "etherpad.png",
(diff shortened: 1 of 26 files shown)


yegortimoshenko:
To this and all other "drop/remove" commits I've opened in the last two days: don't mind the commit. Please comment on what you think about the change and I'll merge manually if everyone agrees this is the way to go (or no one replies :-).

Zegnat:
Haven’t used Cryptpad so I am going to trust your statement that it is comparable. If it is, by all means, drop Etherpad!

This is one of those points where it is always important to ask “how does this compare to the already listed options?” If the answer is that one is straight up an improvement, the other one should be dropped to keep the list lean.

yegortimoshenko:
Here is a demo: https://cryptpad.fr/pad/#/2/pad/edit/U4leXFVN1dUsE1fzvSQ6vT4s/

yegortimoshenko:
All that said, I do like Etherpad's UI much more :-(

Issue #1998, opened 2018-06-01
OnionBrowser is not free anymore


yegortimoshenko:
https://github.com/OnionBrowser/OnionBrowser#onion-browser

Basically, one can't distribute it anymore.

PR #2005, opened 2018-06-02
Drop Onion Browser, see #1998

yegortimoshenko:
See #1998. Onion Browser is nonfree (distribution in binary form is not allowed), but source code is available. I think we should replace it with some other Tor browser for iOS (skimming through GitHub search, I found at least one).

I feel that nonfree apps with source availability but limited distribution are an interesting corner case for this list, so please let's discuss. cc @Atavic @Hillside502 @quantumpacket @strugee @Zegnat

### DIFF
--- a/source/db/ar-projects.json
+++ b/source/db/ar-projects.json
@@ -3466,31 +3466,6 @@
   },
   {
     "development_stage": "released",
-    "description": "Surf the web through the Tor network with this open source browser for iOS devices.",
-    "license_url": "https://github.com/mtigas/iOS-OnionBrowser/blob/master/LICENSE",
-    "logo": "onion-browser.png",
-    "privacy_url": "",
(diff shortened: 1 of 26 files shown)


quantumpacket:
I suspect the license change may be in part to do with this:

> Copies of Onion Browser downloaded outside the App Store — via app directories like Cydia or other "jailbreak" app directories — are not created by the developer of Onion Browser. These copies may have modifications added by the packager, which may interfere with your privacy or security, or they may secretly contain malware. We cannot provide support for copies of Onion Browser acquired through these channels.

That being the case, I can't fault them for that. However, limiting the routes for binary installation would hinder adoption, especially if users don't want to use the App Store or don't have the knowhow to compile. We want privacy to be within reach for the common user, right?

:+1: for an alternative Tor browser for iOS if one passes our standards, otherwise keep it until one comes along.

